PR #3, opened 2019-07-18
Add documentation about BLT commit message default rules 

alexsoble:


README.md
```diff
@@ -175,9 +175,13 @@ blt sync
 # Deploy
 
 ```
-ACQUIA_CLOUD_REMOTE_GIT=acquia_git_destination blt artifact:deploy --commit-msg "message here" --branch "branch name here" --no-interaction
+ACQUIA_CLOUD_REMOTE_GIT=acquia_git_destination blt artifact:deploy --commit-msg "BLT-001: Commit message here." --branch "branch name here" --no-interaction
 ```
 
+Note that by default, commit messages need to conform to a strict pattern specified in `build.yml` under `git > commit-msg > pattern`. The default regex pattern is as follows, with "project.prefix" being "BLT" by default:
```
LindsayYoung: Can we also add that it needs to end with a period, since that wasn't intuitive to me?

alexsoble: @LindsayYoung Good idea, done!

PR #9, opened 2019-07-24
Deploying modules

LindsayYoung:
This has the initial config with our "Bunny" webform.

In your local run `drush cim` to import the configuration.


config/default/system.site.yml
```diff
@@ -0,0 +1,14 @@
+uuid: 3bb8cee4-7729-4ef8-acb0-46ca31a0d55d
+name: 'BLTed 10'
```
alexsoble: @LindsayYoung Do we want to change the site name here?

alexsoble: @LindsayYoung Resolved in 03d5575166de8ef47518cff21f1b870e24f1b986.

alexsoble:
This works for me locally, I also don't see any sensitive credentials in the YAML.

PR #11, opened 2019-07-25
Switch to new drush site config approach

alexsoble:
# Notes

+ Environment variable interpolation isn't possible for YML files that are parsed by `drush`
+ Remove the whole `drush/sites/dojportal-blt.site.yml` file from source control so that we can drop in versions with SSH URLs locally and not have to support YML variable interpolation


alexsoble:
This is working locally on @LindsayYoung's machine!!

PR #12, opened 2019-07-25
WIP: Create builds from config

LindsayYoung:
This should be a deployment improvement. We can configure the commands here, so we don't have to run separate drush commands. 

Thanks Marvin!

I committed the existing file but only made 2 small changes. I haven't tested it out yet.

### DIFF
--- a/vendor/acquia/blt/config/build.yml
+++ b/vendor/acquia/blt/config/build.yml
@@ -137,7 +137,7 @@ sync:
   # By default, files are not synced during sync:refresh.
   # Set this value to 'true' or pass -D sync.files=true
   # to override this behavior.
-  files: false
+  files: true
   # Paths to exclude during file syncing operations.
   exclude-paths:
     - styles
--- a/vendor/acquia/blt/config/build.yml
+++ b/vendor/acquia/blt/config/build.yml
@@ -1,0 +1,275 @@
+# @todo Move to subkey of tests.
+behat:
+  config: ${repo.root}/tests/behat/local.yml
+  profile: local
+  # An array of paths with behat tests that should be executed.
+  paths:
+    # - ${docroot}/modules
+    # - ${docroot}/profiles
+    - ${repo.root}/tests/behat
+  tags: '~ajax&&~experimental&&~lightningextension'
+  extra: ''
+  validate: true
+  # May be chrome, selenium.
+  web-driver: chrome
+
+bin.path: vendor/bin
+
+blt:
+  update:
+    ignore-existing-file: ${blt.root}/scripts/blt/ignore-existing.txt
+  config-files:
+    project: ${repo.root}/blt/blt.yml
+    local: ${repo.root}/blt/local.blt.yml
+    example-local: ${repo.root}/blt/example.local.blt.yml
+    schema-version: ${repo.root}/blt/.schema_version
+  command-cache-dir: ${blt.root}/cache/commands
+
+composer:
+  bin: ${repo.root}/${bin.path}
+  extra: ''
+
+cm:
+  # Possible values: core-only, config-split, features, none.
+  strategy: config-split
+  core:
+    # The parent directory for configuration directories, relative to the docroot.
+    path: ../config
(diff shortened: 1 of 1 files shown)


alexsoble:
@LindsayYoung Hmm, the `/vendor` file is gitignored, correct? Since that's the case, how are we committing this file?

LindsayYoung:
Yes, we only want to commit this file with the config.

vendor/acquia/blt/config/build.yml
```diff
@@ -0,0 +1,275 @@
...
+    # E.g., $config_directories['sync']. It must have a corresponding key in cm.core.dirs. E.g., `cm.core.dirs.sync`.
+    key: sync
+    dirs:
+      # Corresponding value is defined in config.settings.php.
+      sync:
+        path: ${cm.core.path}/default
+    # Install site directly from existing config.
+    # This cannot be used if your install profile implements hook_install.
+    # Import our project config
+    install_from_config: true
```
LindsayYoung: added this to import our project config

alexsoble:
@LindsayYoung Ah, I see ...

```
git add --force my/ignore/file.foo
```

https://stackoverflow.com/questions/8006393/force-add-despite-the-gitignore-file

vendor/acquia/blt/config/build.yml
alexsoble: @LindsayYoung Looking at what this does in the BLT code ... 👀 

https://github.com/acquia/blt/blob/46203b6c9c55ef142c9bd9150c7e6d2a14bfd5a0/src/Robo/Commands/Setup/DrupalCommand.php#L61

LindsayYoung:
Tried this on Friday morning and it did not deploy the config. We are probably missing a config the one we set is only for initializing drupal

vendor/acquia/blt/config/build.yml
```diff
@@ -0,0 +1,275 @@
...
+  strategy: install
+  # If setup.strategy is import, this file will be imported. File path is
+  # relative to drupal docroot directory.
+  dump-file: null
+
+sync:
+  # By default, files are not synced during sync:refresh.
+  # Set this value to 'true' or pass -D sync.files=true
+  # to override this behavior.
+  files: false
```
LindsayYoung: I think this needs to be true too, I am going to try it.

alexsoble: Ooh, good idea!

vendor/acquia/blt/config/build.yml
```diff
@@ -0,0 +1,275 @@
...
+  strategy: install
+  # If setup.strategy is import, this file will be imported. File path is
+  # relative to drupal docroot directory.
+  dump-file: null
+
+sync:
+  # By default, files are not synced during sync:refresh.
+  # Set this value to 'true' or pass -D sync.files=true
+  # to override this behavior.
+  files: true
```
LindsayYoung: Added this too, though looks like we are missing something, to get to the point where we don't also need to run the drush commands.

alexsoble: Hmmmmm, let's check in with Marvin from Acquia on Monday, he should know!

alexsoble: I looked to see if there was anything we could borrow from https://github.com/usdoj/foia-api but that repo doesn't have a `vendor` directory so they must not have gone the force-git-adding route.

LindsayYoung:
Did get the site name to update on staging but not dev, I think there might be more settings before we don't need the drush commands. We can check with Marvin on Monday

vendor/acquia/blt/config/build.yml
```diff
@@ -0,0 +1,275 @@
...
+    - files.php.custom.themes
+    - files.frontend.custom.themes
+
+project:
+  local:
+    uri: ${project.local.protocol}://${project.local.hostname}
+
+setup:
+  # Valid values are install, sync, import.
+  strategy: install
```
LindsayYoung: change to sync

LindsayYoung:
closing this in favor of #16

PR #16, opened 2019-07-31
Remove BLT commit message validation; remove git URL config via env

alexsoble:
# Notes

+ Hardcodes git URL as proposed by @marvoey. 
+ Removes overzealous BLT commit message validation. 

# Review steps 

+ [x] Verify deploy from local machine on this branch
  + [x] Verify that deploy will no longer fail if commit message does not match BLT git hook rules


alexsoble:
Tried this locally with ` blt artifact:deploy --commit-msg "Deploy config to dev and staging." --branch "master" --no-interaction --verbose` and continued to see dangerous rsync behavior attempting (unsuccessfully) to delete every file and folder on my local machine. Will file an issue against the acquia BLT repo as suggested by @marvoey.

alexsoble:
@LindsayYoung @marvoey I think over the course of our technical discovery today we've found that the deploy issues we're seeing were related to a missing `vendor/acquia/blt/config` directory, possibly due to composer install issues, and not to anything changed in this PR. 

However, this PR turns off the stringent BLT commit hooks, which is nice. And it hardcodes the git URL, which we might as well implement since there's no point in keeping it environmentalized once it's been committed to git. 

@LindsayYoung, mind giving this a quick review and then let's aim to merge it in?

PR #22, opened 2019-08-05
Apply USWDS styling to radio buttons and checkboxes

alexsoble:
# Screenshots

## Radio buttons

usa-radio-buttons

## Checkboxes

usa-checkmarks

# Thank yous! 

Thank yous to @brianseek, @fureigh, and @brockfanning for help and advice on this pull request!


alexsoble:
Problem with the preprocess hooks in e6b10f886170f103e6900258aeec5fe0692448bb:

They lead to a sledgehammer, "make everything a radio button" approach, which breaks the checkbox questions:

messed-up-checkboxes

alexsoble:
Looking at the HTML output, I'm seeing twig templates rendered by `classy`, `webform`, and `uswds` themes. I know that `uswds` and `webform` should be there, but I'm unclear about the role of `classy`. 

I am wondering about the following:

1. Is the `classy` theme required for `webform` to work? 
   a. How do I find out?
  b. If it is not required by `webform`, can I safely remove it? 

2. The top-level HTML element rendered by `classy` for a set of radio buttons looks like this:

```
<!-- THEME DEBUG -->
  | <!-- THEME HOOK: 'radios' -->
  | <!-- BEGIN OUTPUT from 'core/themes/classy/templates/form/radios.html.twig' -->
  | <div id="edit-primary-complaint-radios" class="js-webform-radios webform-options-display-one-column form-radios">
```

How can I make the preprocessor for labels dynamic, so that if it is enclosed by a `radios` HTML element, it applies a `usa-radio__label` class, whereas if it is enclosed by a `checkboxes` HTML element, it applies a `usa-checkbox__label` class?

fureigh:
@alexsoble What markup does `core/themes/classy/templates/form/radios.html.twig` contain? (I'm cloning the repo, but getting the build up and running will take a few.)

I ask because you may be able to just copy it to a corresponding directory structure in your subtheme and add or modify the class in the markup there instead.

fureigh:
@alexsoble Leave `classy` installed. `themes/contrib/uswds/uswds.info.yml` shows that it's the USWDS theme's base theme, so the USWDS theme won't work without it.

(If the USWDS theme _weren't_ using Classy, I'd say you shouldn't "remove" it per se — Classy is a [core theme](https://www.drupal.org/docs/8/core/themes/classy-theme), and rule number one is don't hack core — but you could uninstall it.)

When you're finished with front-end debugging, though, and on any version being pushed to production, check the settings in `admin/appearance` (in the UI)... those comments in the markup come from debug mode.

alexsoble:
@fureigh Here are the contents of `core/themes/classy/templates/form/radios.html.twig `:

```
{#
/**
 * @file
 * Theme override for a 'radios' #type form element.
 *
 * Available variables
 * - attributes: A list of HTML attributes for the wrapper element.
 * - children: The rendered radios.
 *
 * @see template_preprocess_radios()
 */
#}
<div{{ attributes.addClass('form-radios') }}>{{ children }}</div>
```

I can give that approach a try! Since two of the three USWDS custom classes are being applied using preprocess hooks, I think it would be better for readability/maintainability to add the third custom class with a preprocess hook as well. Maybe I can add a radio button preprocess hook that adds this class to all radio button elements? I'll give that a try too.

alexsoble:
@fureigh "Don't hack core!" Good rule!

fureigh:
@alexsoble Great! Simpler yet: have you looked at `template_preprocess_radios()`?

fureigh:
@alexsoble ^ I recommend that approach instead. :)

https://api.drupal.org/api/drupal/core%21includes%21form.inc/function/template_preprocess_radios/8.2.x

fureigh:
(...so it sounds like we're on the same page! 🍀)

brockfanning:
To me this seems like a shortcoming in the Drupal theme. All radio buttons should have this class right? We should try to get it added to the Drupal theme, I think, once you've got it working here. [Here](https://git.drupalcode.org/project/uswds/blob/8.x-2.x/preprocess/element/input__textfield.preprocess.inc) is another example of adding a class in a preprocess function. You could try something similar in the subtheme, only you'd name the function `mysubtheme_preprocess_radios` (adjusted according to your subtheme name).

alexsoble:
Thank you @fureigh and @brockfanning!! Brock, agreed that this is a shortcoming in the Drupal theme. @fureigh and I are planning to file an issue, though I'm currently stuck with an error when registering for a Drupal.or account.

alexsoble:
@fureigh @brockfanning I tried the radio button (and checkbox) preprocessing approach in eb4326dd227492553455989db45787884fadbac2 and e53ae2c365ceae061f1ff6bc6bd80a516f6b2361, with different function naming conventions:

```
* crt_portal_subtheme_preprocess_radios
* crt_portal_subtheme_preprocess_input__radio
```

One naming style matches `template_preprocess_radios()`, the other matches something I'm seeing in the HTML output: 

```
<!-- THEME HOOK: 'input__radio' -->
```

So far neither is applying the class correctly. I am running `drush cr` in between updates. 

Do I need to adjust the file names of the preprocess hooks themselves? My assumption is that the function names are what matter here and the filenames can be arbitrary, but that might be an incorrect assumption!

brockfanning:
@alexsoble Yes I think the the filenames are important. See some docs [here](https://git.drupalcode.org/project/uswds/blob/8.x-2.x/examples/my_subtheme/preprocess/README.md). You might first try your original approach but with the files called "checkboxes.preprocess.inc" and "radios.preprocess.inc".

alexsoble:
In line with the adding-another-preprocess-hook strategy, @brianseek graciously shared an `input.preprocess.inc` file that does the trick: f9239983fc7202b71fc3fff4e969e4c015630095.

alexsoble:
@brockfanning I tested out the `checkboxes.preprocess.inc` and `radios.preprocess.inc` suggestion, unfortunately changing those filenames did not apply the desired classes.

I wonder if there's a documented list somewhere that shows the filenames required for each HTML element's preprocess hook? Or does it vary by theme somehow? 

Since `input.preprocess.inc` is working, I won't dig too much further here. But if you know of such documentation or find it down the road, I would be all ears! 👂

brockfanning:
Nice, that input.preprocess.inc looks good.

For future reference this is the way it works: The filename depends on the "hook" in the preprocess function you're trying to use. This is the part after "mytheme_preprocess_". So if the function is mytheme_preprocess_foo, then the "hook" is foo. The filename is supposed to be [hook].preprocess.inc. So for the example above it would be a file called foo.preprocess.inc, containing a function named mytheme_preprocess_foo.

The only quirk is with dashes/underscores. A rule of thumb is that in function names, use underscores, and in file names, use dashes. For example, if the "hook" is "foo_bar", then you would need a file named foo-bar.preprocess.inc, containing a function named mytheme_preprocess_foo_bar.

The README in the preprocess folder was intended be this documentation, but let us know on drupal.org if you think it needs improvement.

That said, that's the way it's _supposed_ to work. I'm not sure why it was not working for you earlier.

alexsoble:
@brockfanning Ah, thank you! Helpful to understand that the filename should match the function name, and thank you for pointing out the issue with dashes/underscores — super helpful.

But for our radio button example: if I'm trying to add a class to all radio buttons, how do I know if the hook should be `radio`, `radios`, `input__radio`, or `input__radios`?

brockfanning:
That may be the point of the "theme debug" setting. So if you see `<!-- THEME HOOK: 'radios' -->` then you know the hook is "radios".

alexsoble:
@brockfanning That makes sense!

LindsayYoung:
FYI @alexsoble  this does require a `drush cim` because of the updates to: 
     core.entity_view_display.node.page.teaser 
     core.entity_view_display.node.page.default
     core.entity_form_display.node.page.default

PR #23, opened 2019-08-07
Assignment logic + updated webform

LindsayYoung:
This adds the great improvements that @Jacklynn has been working on and implements the voting or not voting logic in issue https://github.com/18F/crt-portal/issues/54

This requires an import.

removes other versions of the form for clarity.


LindsayYoung:
In corporates https://github.com/18F/crt-portal/issues/94

Also here is the Twig syntax cheat sheet https://danielmg.org/php/twig-cheatsheet.html

LindsayYoung:
Moving repos

PR #26, opened 2019-08-19
Add form wizard progress bar and progress text

alexsoble:
## Screenshot (visual progress bar)

header-area

## Audio recording (screen reader progress text)

This PR adds screen-reader-only progress text to accompany the visual progress bar. 

🔉🔉
 
To listen to the progress text as read by Voice Over + Safari on Mac, download and play [this mp3 recording](https://raw.githubusercontent.com/usdoj/crt-portal/ars/wizard-progress-bar/audio_testing/wizard-progressbar-audio-test-8-20-2019.mp3). (GitHub's content security policy does not permit directly embedding audio files.)

🔉🔉

## Notes

Wizard progress bar CSS code adapted from Carlos Ballena via CodePen (MIT License). 

https://codepen.io/cballenar/pen/vxpZxE


alexsoble:
Thinking through two choices at this point.

## Progress bar steps as links — high-priority or skippable for now? 

As I recall, the Drupal wizard library made each previous progress bar step a link to return to an earlier section of the form.

Re-building this functionality within the framework of the Django form wizard library might be possible, but isn't straightforward out of the box. The library offers [examples](https://django-formtools.readthedocs.io/en/latest/wizard.html#creating-templates-for-the-forms) of buttons that return the user to previous steps (`wizard_goto_step`). However, these are buttons within the form, not links — so I'd need to adjust the progress bar code to use buttons instead of links to make this work. 

The Django form wizard library does offer users navigation buttons at the bottom of each form section out-of-the-box. The buttons include options to go to "first step, "prev step," or "next step." I wonder if this is enough navigation to meet user needs for now? In that case, could we skip making the progress bar steps into links for now?

alexsoble:
## Incorporate design element(s) from CodePen progress bar?

The CodePen CSS I'm adapting for this progress bar includes some cute design elements, like a dotted line between the current step and the next step: https://codepen.io/cballenar/pen/vxpZxE — would we like to incorporate any of these into our own progress bar?

alexsoble:
Good feedback from @uchmok and @avivaoskow in standup — we can always iterate on the progress bar later, let's aim to ship this PR without adding more functionality to it. I agree, will focus in that direction!

crt_portal/cts_forms/templates/forms/report.html
```diff
@@ -3,27 +3,61 @@
 {% block head %}
 {{ wizard.form.media }}
 {% endblock %}
-
 {% block content %}
+  <span class="usa-sr-only">
```
LindsayYoung: I am thinking if we create the text in the view and pass these in to the template, it will make internationalization easier.

Here is some background on how to do that in Django: https://docs.djangoproject.com/en/2.2/topics/i18n/translation/

Translation is not in the thin slice but, I like to make it easier on us or others when that happens. Happy work on it with you if you want; What do you think?

alexsoble: Thanks for the note! I'll keep this in mind going forward. And I would love to work on this together during an upcoming pairing session. Since translation is not in the thin slice of our current scope, I would be comfortable shipping this current PR without translation support; we can add translation support for this feature in a future PR.

LindsayYoung: Sounds good, to be clear, we would not translate it now either way, I was only suggesting to add the logic to the view and pass the variables to the template. But it sounds like we are on the same page with it not being necessary for this PR.

crt_portal/cts_forms/templates/forms/report.html
```diff
@@ -3,27 +3,61 @@
...
+    {% elif wizard.steps.step1 == 2 %}
+      Where
+    {% elif wizard.steps.step1 == 3 %}
+      Who
+    {% elif wizard.steps.step1 == 4 %}
+      Details
+    {% elif wizard.steps.step1 == 5 %}
+      Contact
+    {% elif wizard.steps.step1 == 6 %}
+      Review & Submit
```
LindsayYoung: Since we have not made the review functionality, should it just be "Submit" for now?

alexsoble: Sure!

alexsoble: Implemented here: c6b7d287b98add15c4183d47fa3badea72ff1d3e.

crt_portal/cts_forms/templates/forms/report.html
alexsoble: Yep!

alexsoble:
@LindsayYoung Added some more styling in 0c32e2b1479f061f1211f21fdf1bdcb9e7982a6a; will update the screenshot to match.

crt_portal/cts_forms/templates/forms/base.html
```diff
@@ -53,19 +53,25 @@
...
+    <header class="page-header-background">
       <div class="grid-container">
         <div class="grid-row">
           <div class="grid-col-1"></div>
           <div class="grid-col-10">
             <h1>
               Contact the Department of Justice about a civil rights concern
             </h1>
-            {% block content %} {% endblock %}
+            <p class="intro-subtitle">
```
LindsayYoung: This is the template base we will want to use for all or many of the pages. So we will want to add a block here- something like `{% block intro %}{%  endblock %}` and then we can add the text on the form page in a corresponding block.

alexsoble: Hmmmm, from what I see of the designs, every single page has the exact same intro text. We could extract this out into a block, but I think that would be a premature optimization.

LindsayYoung: We know we will need a homepage and, generally, we want to make software that is easy to extend. I would agree with you about it being premature if it was more work to implement in a way that made it extensible, but since it is a wash, let's get that in now.

alexsoble: OK, sounds good. Done in e78fbecfc4a1dfab94a4a6e60b0d8b1c0fd53657. I extracted out the whole header area so that the h1 title, intro subtitle, and other details of the header can be customized from page to page.

alexsoble:
@LindsayYoung 🙌

alexsoble:
@LindsayYoung Thank you for the review!

PR #27, opened 2019-08-21
Static assets

LindsayYoung:
This configures static assets so that they work on cloud.gov. 

And look, they do https://crt-portal-django-prod.app.cloud.gov/report/

You might need to rebuild the docker image locally if the static assets don't show up right away.

closes https://github.com/18F/crt-portal/issues/105


.cfignore
```diff
@@ -0,0 +1,2 @@
+docker-compose.yml
+Dockerfile
```
alexsoble: The [CloudFoundry docs on .cfignore](https://docs.cloudfoundry.org/devguide/deploy-apps/prepare-to-deploy.html#exclude) mentions excluding `/tmp` and `/log` directories; are there any temporary or log directories we'd want to add here? `node_modules`?

Dockerfile
```diff
@@ -11,6 +11,8 @@ WORKDIR /code
 
 # Install Python dependencies
 RUN pip install pipenv
+
```
alexsoble: Whitespace diff

Pipfile.lock
```diff
@@ -18,18 +18,18 @@
     "default": {
```
alexsoble: The Pipfile itself didn't change, so it looks like these changes are unrelated to the main work of this PR.

README.md
```diff
@@ -42,29 +50,53 @@ in another terminal if you are doing front end work:
...
-(I had to give it the varable name in this command and follow up with the secret key, I am not sure why)
+    cf push
+
+That will push to cloud.gov according to the instructions in the manifest and Profile.
+
+### Create admin accounts
 
 Need to ssh to create superuser (would like to do this automatically in another PR)
 
     cf ssh crt-portal-django
```
alexsoble: I haven't set up CloudFoundry locally yet, would like to go through some of these steps as part of review.

README.md
```diff
@@ -42,29 +50,53 @@ in another terminal if you are doing front end work:
...
+Then, you can create a superuser
+
+    python /crt_portal/manage.py createsuperuser
 
-Needed to enable CORS via awscli https://cloud.gov/docs/services/s3/#allowing-client-side-web-access-from-external-applications (would like to do this automatically in another PR)
+### Subsequent deploys
 
+Once cloud.gov is set up, you can deploy just with a push
+
+    cf push
```
alexsoble: Review step — push a commit to prod!

.cfignore
LindsayYoung: The Node modules are getting pushed to cloud.gov. The static files generated by the Node modules are going to be copied to S3 with the `collectstatic` command but the node modules are with the code in cloud.gov. In the future, we will build the node modules in Circle, that is when we would add it to the .cfignore. 

I am not seeing a tmp or a log directory, so I don't think we need it but no harm in adding it.

README.md
alexsoble: Configured locally during pairing time today!

README.md
alexsoble: Done during this morning's pairing, thanks @LindsayYoung!

run.sh
```diff
@@ -3,6 +3,9 @@
 echo migrate database...
 python /code/crt_portal/manage.py migrate
 
+echo collect static assets...
+python /code/crt_portal/manage.py collectstatic --noinput
```
alexsoble: Does this line mean that production and local envs are serving static files in the same way? I was thinking this PR was only changing the production static assets path, but that might be wrong.

crt_portal/crt_portal/settings.py
```diff
@@ -158,11 +156,13 @@
     STATIC_URL = f'https://{AWS_S3_CUSTOM_DOMAIN}/{AWS_LOCATION}/'
     STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
-
 else:
     STATIC_URL = '/static/'
 
-STATIC_ROOT = os.path.join(BASE_DIR, "static")
+STATICFILES_DIRS = ( os.path.join(BASE_DIR, 'static'), )
+ADMIN_MEDIA_PREFIX = STATIC_URL + 'admin/'
```
alexsoble: @LindsayYoung It looks like `ADMIN_MEDIA_PREFIX` is deprecated, does that match your understanding? https://docs.djangoproject.com/en/dev/releases/1.4/#django-contrib-admin

README.md
```diff
@@ -30,9 +30,17 @@ in another terminal if you are doing front end work:
 
     gulp watch
 
-
+Also note, that the staticfiles folder is the destination of all static assets when you or a script runs `manage.py collectstatic` so don't make your changes there, or they will be overwritten.
 
 ## cloud.gov set up
+You only need to get the services stood up, configure the S3 bucket once.
+
+For working with cloud.gov directly, you will need to [install the cloud foundry cli](https://docs.cloudfoundry.org/cf-cli/install-go-cli.html). That will allow you ro run the `cf` commands in a terminal.
```
alexsoble: Typo here: "ro run".

README.md
```diff
@@ -30,9 +30,17 @@ in another terminal if you are doing front end work:
 
     gulp watch
 
-
+Also note, that the staticfiles folder is the destination of all static assets when you or a script runs `manage.py collectstatic` so don't make your changes there, or they will be overwritten.
 
 ## cloud.gov set up
+You only need to get the services stood up, configure the S3 bucket once.
```
alexsoble: This might be clearer as: "You only need to get the services stood up and configure the S3 bucket once."

Pipfile.lock
alexsoble: @LindsayYoung Let's back out changes to the Pipfile.lock since they're unrelated, does that sound good to you?

Dockerfile
alexsoble: There's still a whitespace diff here, not a huge deal but just something we can catch in `git status` / `git diff head` checks.

.cfignore
alexsoble: This is helpful context! I was asking about the `/tmp` and `log` directories in an open-question way. We certainly don't need to add those lines to the .cfignore file if those directories do not exist.

alexsoble:
@LindsayYoung I think we should change the title of this PR to reflect that in addition to setting up static assets, it unblocks deploy to cloud.gov and documents how to deploy! Lots of great stuff here!

LindsayYoung:
That is how the ticket is named- and that is what is up on the zenhub board- rename it if you want to, and merge it if you are done reviewing. Thanks again @alexsoble

run.sh
LindsayYoung: It will now copy your files locally to /staticfiles, in cloud.gov it will copy to S3, This will make sure that collect static works similarly, in the local environment and doesn't give us surprises in production.

Pipfile.lock
LindsayYoung: I wanted to add the awscli as a dev dependency, but I did not realize it was already a dev dependency,(Thanks past me), and I just updated them when I re-added the packages. 

I would like to merge the lock because the updated packages are what you and I deployed.

But, thank you for catching this.

alexsoble: OK, that makes sense to me!

.cfignore
alexsoble: In fact — let's remove `/tmp` and `/log` from here since they don't exist.

run.sh
alexsoble: Gotcha, thank you!

crt_portal/crt_portal/settings.py
```diff
@@ -158,11 +156,12 @@
     STATIC_URL = f'https://{AWS_S3_CUSTOM_DOMAIN}/{AWS_LOCATION}/'
     STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
-
 else:
     STATIC_URL = '/static/'
 
-STATIC_ROOT = os.path.join(BASE_DIR, "static")
+STATICFILES_DIRS = ( os.path.join(BASE_DIR, 'static'), )
+MEDIA_URL = 'media/'
```
alexsoble: @LindsayYoung Trying to figure out what this MEDIA_URL variable is doing — seems like it has to do with [file uploads](https://docs.djangoproject.com/en/2.2/ref/settings/#media-url). My understanding is that we don't have any features involving file uploads in the thin slice; can we safely remove this line?

crt_portal/crt_portal/settings.py
```diff
@@ -158,11 +156,12 @@
...
+STATIC_ROOT = os.path.join(BASE_DIR, "staticfiles")
```
alexsoble: @LindsayYoung Would you mind adding one or two lines of code comments about how STATIC_ROOT and STATICFILES_DIRS work? Trying to wrap my mind around how Django static assets are working here. Thank you!

LindsayYoung: I did add the one liners- Static assets are confusing I added some one liners. There is a link to the docs above this section. 

I would recommend watching our former college, Jacob: 
https://www.youtube.com/watch?v=E613X3RBegI

alexsoble: Thanks!

PR #24, opened 2019-08-15
Add more complex models

LindsayYoung:
Still working on this. So far, the single answer multi select fields are easy. I need to make modification for the forms.py and models for multi select fields.

closes https://github.com/18F/crt-portal/issues/104

PR #28, opened 2019-08-21
Add commands to readme

LindsayYoung:
I realized when @ilodidoj asked for the commands to ssh,  that I had made this, adding this as a PR now. 

It explains the commands to do data migrations via Docker and to ssh into the Docker container. The .env file is useful for ssh.


README.md
```diff
@@ -30,6 +30,18 @@ in another terminal if you are doing front end work:
 
     gulp watch
 
+## Running common tasks
+
+When running migrations, make sure you set up a SECRET_KEY in a .env file for yourself locally
```
alexsoble: Is the `.env` file loaded automatically by the Django app?

LindsayYoung: The .env setting is for having a variables available in docker during the pre-processing step, so having the secret key set there helps with running commands via docker. Variables in the env file are not available to the app after the build.

README.md
```diff
@@ -30,6 +30,18 @@ in another terminal if you are doing front end work:
...
+
+In Django, when you update the data models you need to create migrations and then apply those migrations, you can do that with:
+
+    docker-compose run web python /code/crt_portal/manage.py createmigrations
```
LindsayYoung: I am seeing a typo here it should be "makemigrations" I can fix that

README.md
alexsoble: Like we talked about — `SECRET_KEY` is supposed to be defined automatically in the local environment by a function in `local_settings.py`, but that is not working for me. The README suggests exporting a secret key manually. And this introduces a third path to setting a secret key. We should consolidate down to a single path for setting a secret key and make sure that works properly in the local environment for all devs.

LindsayYoung:
@alexsoble this has the secret key simplifications that we talked about.

alexsoble:
@LindsayYoung OK, great! I'm seeing this PR keeps the code in `crt_portal/local_settings.py` that also sets SECRET_KEY; I'm wondering what the relationship between those two secret keys is going to be? 

I also recognize that the main point of this PR is to add documentation on how to run migration commands and not necessarily to change the SECRET_KEY setup, so I'd be happy if we address the SECRET_KEY issue in a different PR (as long as the total number of different code paths setting SECRET_KEY doesn't increase here).

LindsayYoung:
We can simplify things now - not a problem just pushed that change @alexsoble

alexsoble:
@LindsayYoung Excellent! I love that we're down to a single place in the code that sets SECRET_KEY; it works for me locally now! 

I see there's a merge conflict (probably since I merged in #24); once that's fixed let's 🚢!

LindsayYoung:
@alexsoble fixed the merge conflict and removed unused imports

alexsoble:
@LindsayYoung Looks great to me, glad we've simplified the SECRET_KEY path! Thanks for this!

PR #40, opened 2019-08-29
Django form style customizations

alexsoble:
Link to ZenHub issues:

+ [Override default Django form wizard HTML; restructure for vertical flow](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/120)
+ [Apply USWDS classes/styling to radio buttons and checkboxes](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/121)

## What does this change?

+ Replaces generated HTML from `wizard.form` with custom form HTML. 
  + This will let us put form question labels above the inputs (like in Sketch) instead of off to the left (like they are in the cloud.gov site right now).
+ Adds USWDS styling to radio buttons.
+ Adds a few initial styling updates to question labels.

### Old
old

### New
new

## Checklist:

### Author 

+ [x] If front end or functionality change, run locally and check http://0.0.0.0:8000/report/.
+ [x] Tests pass locally: `docker-compose run web python /code/crt_portal/manage.py test cts_forms`.
+ [x] pa11y manual check against the `/report` page returns no errors.

### Reviewer 

+ [ ] Tests pass locally: `docker-compose run web python /code/crt_portal/manage.py test cts_forms`.
+ [ ] pa11y manual check against the `/report` page returns no errors.

## Notes for reviewer:

+ I found this blog post on customizing Django form HTML super helpful in tackling this issue: [How to Render Django Form Manually](https://simpleisbetterthancomplex.com/article/2017/08/19/how-to-render-django-form-manually.html) by Vitor Freitas.


alexsoble:
@LindsayYoung Ready for review when you're ready! 😄

alexsoble:
☝️ Force pushed to remove a commit that I meant to put on a separate branch/PR: #41.

alexsoble:
☝️ Changed my mind, will also include question title and helptext styles on this PR so that there are fewer PRs to review on Tuesday.

LindsayYoung:
Following up from our Slack conversation and posting here for to document the conversaion, here is an example of extending the RadioSelect to apply a custom class. https://github.com/usdoj/crt-portal/blob/class-test/crt_portal/cts_forms/forms.py#L11 

It looks like it will be less code. Tell me what you think, @alexsoble. Thanks!

alexsoble:
@LindsayYoung Here's what I see when I run the suggested code on the `class-test` branch locally, do you see the same thing when you check locally?

no-buttons

LindsayYoung:
Sorry, that wasn't supposed to be stand alone code- it was only the part of the code that adds class, you can see it has the class in inspector.

alexsoble:
@LindsayYoung Gotcha! My thinking here is that I already had to set a custom `option_template` HTML template (in this case, `usa_radio_option.html`) in order to set a `usa-radio__label` custom class on the `<label>` element. 

From there, it was only a single line of code to add the `usa-radio__input` class to the input element. I think it makes sense to apply all custom classes to element in the same HTML file, rather than setting some classes via HTML and setting others via Python using `self.attrs`. 

Let me know if that logic and decision-making makes sense to you. Thank you!

LindsayYoung:
We can also add the label and the custom label and html class by extending the form class. So we don't have to overwrite the html if we don't want to. 

(Thanks for you patience on this reply, I thought I sent it yesterday but it was sitting it a tab.)

alexsoble:
Hey @LindsayYoung! It would be excellent if we could add the custom label class without overriding the HTML template. Unfortunately, according to what I've been able to discover online, overriding the HTML template is currently the way to add custom classes to Django-generated form labels:

* https://code.djangoproject.com/ticket/29153
* https://stackoverflow.com/questions/6959178/how-to-set-css-class-of-a-label-in-a-django-form-declaration

Please take a look at the above links when you have a moment; if I'm missing something here, let me know! Thanks.

LindsayYoung:
Thanks for the clarification!

PR #43, opened 2019-08-30
Add .gitattributes file to enforce LF line endings

alexsoble:
## What does this change?

Adds a `.gitattributes` file to enforce consistent line ending behavior for all users across Windows and Mac machines.

This should help remove a class of issues related to auto-conversion of line endings that we're running into as we debug @ilodidoj's local setup.

More: https://help.github.com/en/articles/configuring-git-to-handle-line-endings


LindsayYoung:
Would you like a review and merge on this one?

alexsoble:
@LindsayYoung Yes, thanks! :D

PR #49, opened 2019-09-06
Add USWDS styling to CheckboxSelectMultiple widget

alexsoble:
## What does this change?

+ Adds USWDS styling classes to checkboxes generated by Django form wizard. 

## Screenshot (for front-end PR):

checkboxes-uswds

## Checklist:

+ [x] If front end or functionality change, run locally and check http://0.0.0.0:8000/report/.
+ [x] pa11y manual check against the `/report` page returns no errors.

## Notes for reviewer:

+ Our last USWDS-styling PR reverted some functionality that we wanted to keep (HTML validation messages on required radio buttons), so let's double-check to make sure this PR doesn't revert any functionality. I checked and didn't find anything, but a second pair of eyes never hurts!

### DIFF
--- a/crt_portal/cts_forms/widgets.py
+++ b/crt_portal/cts_forms/widgets.py
@@ -5,3 +5,29 @@ class UsaRadioSelect(ChoiceWidget):
     input_type = 'radio'
     template_name = 'django/forms/widgets/radio.html'
     option_template_name = '../templates/forms/widgets/usa_radio_option.html'
+
+
+class UsaCheckboxSelectMultiple(ChoiceWidget):
+    allow_multiple_selected = True
+    input_type = 'checkbox'
+    template_name = 'django/forms/widgets/checkbox_select.html'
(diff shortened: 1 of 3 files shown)


crt_portal/cts_forms/widgets.py
```diff
@@ -5,3 +5,31 @@ class UsaRadioSelect(ChoiceWidget):
     input_type = 'radio'
     template_name = 'django/forms/widgets/radio.html'
     option_template_name = '../templates/forms/widgets/usa_radio_option.html'
+
+
+# Overrides Django CheckboxSelectMultiple:
+# https://docs.djangoproject.com/en/2.2/ref/forms/widgets/#checkboxselectmultiple
+class UsaCheckboxSelectMultiple(ChoiceWidget):
```
alexsoble: Could be an improvement if this inherits from `CheckboxSelectMultiple` instead of `ChoiceWidget`...

PR #52, opened 2019-09-10
Contact page

LindsayYoung:
[Contact page](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/35)

## What does this change?
Fist pass at the contact form. This is now the first page and includes first name, last name, email and home. This should have some validation on phone.
to come
## Checklist:

+ [ ] Tests pass locally: `docker-compose run web python /code/crt_portal/manage.py test cts_forms`.
+ [ ] Running `flake8` in root returns no style errors.
+ [ ] pa11y manual check against the `/report` page returns no errors.


LindsayYoung:
- add additional text
- change back from a button to a link
- get clarity on phone validation

alexsoble:
Picking this up from @LindsayYoung to work on styling, a small a11y fix, and other things we need to get this over the finish line!

crt_portal/cts_forms/question_group.py
```diff
@@ -1,5 +1,8 @@
 from django.forms.forms import BoundField
 
+# Adapted from this blog post by Michael Kowalchik (mikepk):
+# https://mikepk.com/2010/08/python-django-forms-errors-fieldsets/
```
LindsayYoung: This is a great blogpost!

alexsoble: Yeah!

crt_portal/cts_forms/tests.py
```diff
@@ -109,3 +87,52 @@ def test_required_where(self):
...
+        form = Contact(data={
+            'contact_first_name': '',
+            'contact_last_name': '',
+            'contact_email': 'foo@bär.com',
+            'contact_phone': ''
+        })
+        self.assertTrue(form.is_valid())
+
+    def test_phone_too_short(self):
+        """Model validation tests need to be structured differently"""
```
alexsoble: @LindsayYoung Can you say more about this code comment? Not sure what it's saying.

LindsayYoung: Would this be better: """Model validation unit tests require testing the model directly"""
I am trying to and an explanation as to why this looks different from the other tests.

alexsoble: Ah, that's helpful! I thought the earlier comment was saying something like, "this needs to be changed sometime down the line."

alexsoble:
## Screenshots of front-end progress to date (Chrome, desktop)

Above the fold:
Screen Shot 2019-09-17 at 11 17 36 AM

Below the fold:
Screen Shot 2019-09-17 at 11 18 44 AM

alexsoble:
☝️ Wooohooo! A passing CI build! 🎈

alexsoble:
Alrighty! Applied 5bb2f65a481b18f7165e7ecbc9ba98ff82519e5d, now the site looks decent on IE11!


ie-11-1

ie-11-2

ie-11-3

crt_portal/cts_forms/phone_regex.py
```diff
@@ -0,0 +1,16 @@
+# This regex is designed with U.S. and international phone numbers in mind.
```
LindsayYoung: We could move this to a test

crt_portal/cts_forms/tests.py
```diff
@@ -109,3 +87,62 @@ def test_required_where(self):
...
+        self.assertTrue(form.is_valid())
+
+    def test_invalid_phone(self):
+        form = Contact(data={
+            'contact_first_name': '',
+            'contact_last_name': '',
+            'contact_email': 'foo@bär.com',
+            'contact_phone': '33333333333333333333333333'
+        })
+        self.assertFalse(form.is_valid())
```
LindsayYoung: I like testing for the correct error. In this case,
` self.assertTrue(phone_error_message == ['Enter a valid value.'])`
because then we are checking for the right error and won't get noise or miss something if there is a different error,

crt_portal/cts_forms/views.py
```diff
@@ -9,9 +9,27 @@
 
 
 class CRTReportWizard(SessionWizardView):
-    """Once all the sub-forms are submitted this class will clean data and save."""
     template_name = 'forms/report.html'
 
+    """Once all the sub-forms are submitted this class will clean data and save."""
```
LindsayYoung: if we want to use http://www.sphinx-doc.org or some other automated doc generation in the future I think this docstring needs to be right under the class. Though, correct me if I am wrong.

crt_portal/cts_forms/phone_regex.py
alexsoble: I'd like to keep a code comment near the regex that explains its intention. 

But testing both US and international phone numbers and calling that out in the comments makes sense to me!

crt_portal/cts_forms/tests.py
alexsoble: There's a test for the correct error here: https://github.com/usdoj-crt/crt-portal/pull/52/files/5bb2f65a481b18f7165e7ecbc9ba98ff82519e5d#diff-18509124e855af21de52f53726ce2005R130.

crt_portal/cts_forms/views.py
alexsoble: I'm sure you are right, feel free to move the docstring if you think that makes sense.

crt_portal/cts_forms/phone_regex.py
LindsayYoung: Ok

alexsoble: Oh wait! Maybe I misunderstood what you were saying ... were you suggesting writing tests for each of the PASS/FAIL cases I listed?

PR #54, opened 2019-09-17
Clean Migrations

LindsayYoung:
This is the clean up step for the migration squish https://github.com/usdoj-crt/crt-portal/pull/45 

Deleted the files that the squished migrations replaced. This should work for people going forward. I waited to do this PR to make sure all environments were in a state where they had run the 34 migrations or hadn't started.

tests pass: https://circleci.com/gh/18F/crt-django/142


alexsoble:
GitHub says: "You’re not authorized to merge this pull request." 😞

alexsoble:
Now I have the right permissions!

PR #61, opened 2019-09-20
Deploy name tweaks

LindsayYoung:
takes care of , most of [finish cloud deploy](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/141)
Once we can confirm this works, we can set the default branch to develop.
## What does this change?

fixes a few naming inconsistencies so that we can deploy from circle to dev and staging

successful deploy to Dev:
https://circleci.com/gh/usdoj-crt/crt-portal/51

successful deploy to Stage:
https://circleci.com/gh/usdoj-crt/crt-portal/56

Merging this to Master will confirm that works too

## Checklist:


.circleci/config.yml
```diff
@@ -118,7 +118,7 @@ jobs:
       - run:
           name: Deploy to dev
           command: |
-              ./blue-green-deploy.sh $USERNAME_DEV $CF_PASSWORD_DEV dev
+              ./blue-green-deploy.sh $CRT_USERNAME_DEV $CRT_PASSWORD_DEV dev
```
alexsoble: Gotcha! So `$CRT_USERNAME_DEV` etc. all need to be set correctly in Circle for the deploy scripts to work properly? I think that would be helpful to note in a code comment.

.circleci/config.yml
```diff
@@ -248,7 +248,7 @@ workflows:
             - build_and_test
           filters:
             branches:
-              only: release/*
+              only: /^release.*/
```
alexsoble: So this matches `release` and `release-1`, but rejects `1-release` and `ars-release` — does that summarize the change well?

LindsayYoung: yea, it is looking for branches starting with "release"

.circleci/config.yml
LindsayYoung: Ok I added a comment to the top of the file and the instructions for adding the vars are already in the readme https://github.com/usdoj-crt/crt-portal#initial-cloudgov-set-up

PR #65, opened 2019-09-24
Politely ask search engines not to index us yet

alexsoble:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/142)

## What does this change?

Adds meta tag to header politely asking search engines not to index us until we've officially launched. (We're using our dev/staging/production sites for testing and QA at the moment.)

### DIFF
--- a/crt_portal/cts_forms/templates/forms/base.html
+++ b/crt_portal/cts_forms/templates/forms/base.html
@@ -4,8 +4,9 @@
  <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    {# Asking search engines not to index our site while we're in development mode. #}
+    <meta name="robots" content="noindex">
     <title>U.S. Department of Justice - Civil Rights Division</title>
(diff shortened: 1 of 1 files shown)


crt_portal/cts_forms/templates/forms/base.html
```diff
@@ -4,8 +4,9 @@
  <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    {# Asking search engines not to index our site while we're in development mode. #}
+    <meta name="robots" content="noindex">
     <title>U.S. Department of Justice - Civil Rights Division</title>
-    <!-- TODO: properly manage static assets -->
```
alexsoble: @LindsayYoung My assumption is that this was a TODO from before we had static assets up and running and is now safe to remove; is that right?

LindsayYoung: yes

PR #74, opened 2019-09-26
Details page functionality

alexsoble:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/69)

## What does this change?

Add detail page what happened. See ACs for Story 69, "Tell DOJ CRT what happened (free form)".

TK

### Author

### Reviewer

## Notes for reviewer:


LindsayYoung:
I think we should do the alert styles in a separate PR. @alexsoble this is ready for your review, when you are back. I can't request though GH since you opened the PR.

crt_portal/cts_forms/templates/forms/report.html
```diff
@@ -92,25 +94,52 @@ <h3 class="margin-bottom-1">
...
-        <button name="wizard_goto_step" type="submit" label="first step" value="{{ wizard.steps.first }}" class="usa-button">{% trans "First" %}</button>
-        <button name="wizard_goto_step" type="submit" label="previous step" value="{{ wizard.steps.prev }}" class="usa-button">{% trans "Back" %}</button>
+        <br>
+        <button name="wizard_goto_step" type="submit" label="previous step" value="{{ wizard.steps.prev }}" class="look-like-link margin-bottom-9">{% trans "Back" %}</button>
       {% endif %}
-      <input type="submit" label="submit" value="{% trans "Next" %}" class="usa-button margin-bottom-9" />
+
     </form>
   </div>
+<!-- move this later -->
```
alexsoble: Just thinking ahead here — I think we'll want to move this to the `/static` folder. 

https://docs.djangoproject.com/en/2.2/howto/static-files/

crt_portal/cts_forms/templates/forms/report.html
```diff
@@ -33,9 +33,7 @@ <h1 class="padding-top-5 margin-0">
             </li>
           {% endfor %}
         </ol>
-        <!-- uncomment when there is more than one step on form
```
alexsoble: Unfortunately the connecting line is all out of sync when we uncomment with only two steps; I was too optimistic when I wrote this code comment! Will back out the uncommenting; I think no line is better than ugly-looking line.

crt_portal/cts_forms/templates/forms/report.html
```diff
@@ -75,7 +77,7 @@ <h3 class="margin-bottom-1">
       {% endif %}
 
       {% if wizard.form.question_groups %}
-        {# Phasing in question grouping so that we can add names and help text to groups of questions #}
+        {# This is for question groups, so that we can add names and help text to groups of questions #}
```
alexsoble: @LindsayYoung I think it would be a good idea to pass all questions through a `QuestionGroup`, even if the group has a single question, so that we can have only one question-HTML-rendering blocks instead of two separate paths for rendering HTML for questions.

crt_portal/cts_forms/templates/forms/report.html
LindsayYoung: sounds good

crt_portal/cts_forms/templates/forms/report.html
LindsayYoung: I think we might need more than one template for these pages. Especially once we have protected class, that page is set up as large buttons.

alexsoble: @LindsayYoung 👍👍

crt_portal/cts_forms/templates/forms/report.html
LindsayYoung: Thank you!!

LindsayYoung:
@alexsoble I think this is in good shape any last notes?

alexsoble:
@LindsayYoung I noticed a couple of edge cases where the word count functionality isn't working quite like we'd want it to. 

I'm seeing that when I use the "delete" button to delete words, the word counter isn't always keeping up — do you see the same thing? A few examples in GIF form below: 

details-backspace-2
details-backspace

alexsoble:
@LindsayYoung I'm also noticing that when I submit my data and then use the "Back" button to go back to the Details page, the word counter says "500 words remaining" even though the content of the text area says "testing testing."

LindsayYoung:
Yea the function is only getting called on key up. Looks like we could make it key up or key down 

It looks like, in chrome key up doesn't work for delete - https://stackoverflow.com/questions/10187431/cant-fire-keypress-event-when-press-delete-key/10187723

LindsayYoung:
Ok, I misread before. Do you have a suggestion on another trigger besides key up that will account for the edge case?

alexsoble:
@LindsayYoung So this is weird — I tried refactoring the JS here: 268149bd5468d0f2b3ad18341c2fe57c56d34d2c — and with this commit the delete key issue seems to go away. I'm not 100% sure why; possibly `addEventListener`? Do you see the same thing?

alexsoble:
I think I see a solution for the back button issue as well!

LindsayYoung:
If I hit the back button and come back to that page there is no text? I just rebuild to confirm the behavior.

alexsoble:
Talked this through with @LindsayYoung — back button functionality isn't currently working as we'd like it to. Did some timeboxed investigation and decided to save fixing it for a future Story. 

Shipping 🚢 this PR now to put the core details page functionality in place, e.g. the new page adn the word counter. 

Next PR will tackle the alert states plus any style or a11y tweaks. Shipping this to the dev site will make it slightly easier to test on Internet Explorer.

PR #82, opened 2019-10-03
Add accessibility plan documentation 

alexsoble:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/139)

## What does this change?

+ [x] Add copy of a11y testing plan
+ [x] Update PR template


LindsayYoung:
I think this is a great start! We will want to see how long the checks take and we may want to do the checks per release instead of per-story if they are taking a long time.

PR #91, opened 2019-10-09
Adding back protected class as many to many field

LindsayYoung:
[protected class](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/127)

## What does this change?
Adds the protected class page

Screen Shot 2019-10-15 at 1 03 25 PM
Screen Shot 2019-10-15 at 1 03 06 PM

### Author

+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests)

### Reviewer

## Notes for reviewer:
Need "other" to stay open for back button. 


LindsayYoung:
The other text is tricker than I thought. I just found multiwidget, I am hoping that makes it easier.
https://kite.com/python/docs/django.forms.MultiWidget

LindsayYoung:
I am seeing a circleci/pally testing issue I am going to fix that

LindsayYoung:
fixing merge conflict, I will ping you @alexsoble when it is ready agian

LindsayYoung:
Ok, @alexsoble this should be ready for you tomorrow. Thanks!

crt_portal/cts_forms/templates/forms/report_class.html
```diff
@@ -0,0 +1,59 @@
...
+
+</fieldset>
+<div class="margin-bottom-5"></div>
+{% endblock %}
+
+{% block page_js %}
+<script>
+var otherbox = document.getElementById("other-class-option");
+otherbox.setAttribute("hidden","");
+var otherCheckLabel = document.getElementsByTagName("LABEL").item(14);
```
alexsoble: I think selecting this element based on ID rather than on ordering would make it less fragile; selecting using ordering means that this JS will break if new protected classes are added or removed.

crt_portal/cts_forms/templates/forms/report_single_questions.html
```diff
@@ -0,0 +1,18 @@
+{% extends "forms/report_base.html" %}
+{% block form_questions %}
+<fieldset class="usa-fieldset">
+
+  <div class="page-subtitle">{{ page_subtitle }}</div>
+  {% for field in wizard.form.visible_fields %}
+    <div class="usa-prose margin-bottom-2">
+      <h3 class="usa-prose">{{ field.label }}</h3>
```
alexsoble: One a11y note here — a piece of feedback we got from a blind user was that `<h3>`s shouldn't be used in form labels because their semantic meaning is relation to section heading. I think I created a `label.big` style for this use case instead.

crt_portal/static/js/word_count.js
```diff
@@ -45,11 +46,29 @@ function updateWordCount (e) {
...
+  if (textAreaElem10.length > 0) {
+      updateWordCount(e, max=10, textAreaElem=document.getElementById(textAreaElem10[0].id));
+  };
 };
+
+document.addEventListener('keyup', listenWordCount);
+
+
+function checkOther(elem){
+  var y = document.querySelectorAll('.' + 'usa-checkbox__input').item(13);
```
alexsoble: Let's avoid using variables named `x` and `y` in favor of named variables — clear variable names will make reading the code easier in the future for us or others.

crt_portal/cts_forms/templates/forms/report_class.html
alexsoble: I'm also hoping we can avoid in-line JS in favor of JS served up via assets; if we have JS running both inline and through the assets it will be trickier to debug as opposed to having all the JS in one place.

LindsayYoung: The ids change depending on the PK of the data, so that isn't going to be consistent across environments. The order has to be consistent, as described by the requirements so that is really more stable in this case. 

I am setting the variables in the template because they have to be dynamically generated.

LindsayYoung:
@alexsoble I can't reproduce your js error, can we check in tomorrow so you can screen share and I can see what renders for you?

LindsayYoung:
Thanks again for your input @alexsoble. 

Made the requested changes and resolved the merge conflict. I left the css I added for making the checkboxes flush with the rest of the page but feel free to tear those out in the follow up style pr if you have a better way of doing that.

crt_portal/static/js/word_count.js
```diff
@@ -30,10 +31,10 @@ function updateWordCount (e) {
       wordLimitAlert.removeAttribute('hidden');
       wordLimitAlert.setAttribute('role', 'alert');
     } else {
-      var wordsRemaining = 500 - words;
+      var wordsRemaining = max - words;
       displayCountElem.innerHTML = wordsRemaining;
       countMessageElem.value = (wordsRemaining === 1) ? 'word remaining' : 'words remaining';
-      textAreaElem.classList.remove('bg-gold-outline');
+      // textAreaElem.classList.remove('bg-gold-outline');
```
alexsoble: It looks like this line was commented out by accident — removing this introduces a bug with the text area alert state border, where once the border is added it stays permanently.

package.json
```diff
@@ -4,8 +4,8 @@
   "description": "crt portal",
   "main": "index.js",
   "scripts": {
-    "test:a11y": "pa11y-ci http://0.0.0.0:8000/report/",
-    "test:a11y-ci": "pa11y-ci http://127.0.0.1:8000/report/"
+    "test:a11y": "pa11y-ci --config .pa11y-local.json",
```
alexsoble: @ilodidoj and I found that this command failed for us locally. We discovered the reason why is because my local environment does not include a `pa11y-screenshots` folder, and pa11y does not create the folder for automatically. Adding a `pa11y-screenshots/.gitkeep` folder will fix this.

crt_portal/static/js/word_count.js
```diff
@@ -7,21 +8,21 @@ var wordCountArea = document.getElementById("word_count_area");
 // Show word count area for JS-enabled browsers:
 if (wordCountArea) { wordCountArea.removeAttribute('hidden'); }
 
-function updateWordCount (e) {
+function updateWordCount (e, max='', textAreaElem='') {
```
alexsoble: Setting `''` as the default for `max` will ask the code in this function to do computations like `500 - ''` and `if (words >= '')`. If we don't have a max value supplied, we can't update word count, so I think throwing a JS error would be an improvement on setting a default value here.

.pa11y-local.json
```diff
@@ -0,0 +1,31 @@
+{
+    "urls": [
+        {
+          "url": "http://0.0.0.0:8000/report",
+          "actions": [
+            "wait for element #id_0-contact_phone to be visible"
+          ],
+          "screenCapture": "pa11y-screenshots/page1.png"
```
alexsoble: We loved the screenshots feature, so awesome! Will save some time taking screenshots when doing local front-end development.

crt_portal/crt_portal/settings.py
```diff
@@ -198,8 +198,8 @@
     ('module', 'compressor_toolkit.precompilers.ES6Compiler'),
     ('css', 'compressor_toolkit.precompilers.SCSSCompiler'),
 )
-# COMPRESS_ENABLED = True
-COMPRESS_OFFLINE = True
+
+COMPRESS_ENABLED = True
```
alexsoble: Nice improvement!

crt_portal/cts_forms/forms.py
```diff
@@ -71,32 +71,53 @@ class Meta:
         ]
 
 
-class WhatHappened(ModelForm):
+def retrieve_or_create_choices():
+    choices = []
+    for choice in PROTECTED_CLASS_CHOICES:
+        try:
+            c = ProtectedClass.objects.get_or_create(protected_class=choice)
```
alexsoble: @ilodidoj and I had questions about the variable names here — because `choices` and `choice` are already variable names, it wasn't immediately clear to us what `c` stands for. Something like `protected_class_objects`? A more semantic variable name here could help readers of the code.

crt_portal/static/js/word_count.js
LindsayYoung: Great catch

crt_portal/cts_forms/model_variables.py
```diff
@@ -12,21 +12,39 @@
...
+    'Sex or gender identity (including gender stereotypes) or pregnancy',
+    'Sexual orientation',
+    'Family, marriage, or parental status',
+    'Military status',
+    'Age',
+    'Genetic information',
+    'Other',
+)
+
+PROTECTED_MODEL_CHOICES = (
```
alexsoble: @ilodidoj and I had a couple of questions here:

* What is the difference between `PROTECTED_CLASS_CHOICES` and `PROTECTED_MODEL_CHOICES`?
* Is there ever a case where we'd want the second tuple item in `PROTECTED_MODEL_CHOICES` to not match `PROTECTED_CLASS_CHOICES`?
* If they should always match, should we derive the values in one constant from the values in the other constant to make sure they are a guaranteed match?

crt_portal/cts_forms/templates/forms/report_details.html
```diff
@@ -2,15 +2,20 @@
...
   {% for field in wizard.form.visible_fields %}
     <div class="usa-prose margin-bottom-2">
       <label for="{{ field.id_for_label }}"
              class="big margin-bottom-0">
         {{ field.label }}
       </label>
       <p><em>{{ field.help_text }}</em></p>
-      {{ field.errors }}
+      {% if field.errors %}
+        <div class="usa-alert usa-alert--warning usa-alert--slim">
```
alexsoble: Could you say more about the reason for adding an alert div here? There is already an alert div on the `report_details.html` page about ~15 lines below, so this may be introducing a duplicative alert.

crt_portal/cts_forms/tests.py
```diff
@@ -55,17 +53,24 @@ def test_Contact_valid(self):
         })
         self.assertTrue(form.is_valid())
 
+    def test_Class_valid(self):
```
alexsoble: Should we add a test for the invalid path as well, to make sure the form doesn't mark all protected class data as valid?

.pa11y-local.json
LindsayYoung: I didn't even think of that use- great point!

crt_portal/cts_forms/model_variables.py
LindsayYoung: @ilodidoj and @alexsoble  The `PROTECTED_CLASS_CHOICES` determine what is populated in the form. `PROTECTED_MODEL_CHOICES` determine what is allowable data in the models. 

For example, let's say there is a new need to track national origin and ethnicity separately. Now we just track "National origin (including ancestry, ethnicity, and language)".

So for the old record, we don't want to change them, because we don't know how people replied to the form, but going forward, we don't want to give "National origin (including ancestry, ethnicity, and language)" as an option on the form anymore. 

To do that, we would keep that and add the new classes "National origin" and "Identity" in `PROTECTED_MODEL_CHOICES` and leave the original option. That way, we don't change any of the old data. To make sure the public facing options are correct, we would remove "National origin (including ancestry, ethnicity, and language)" and add "National origin" and "Identity" in `PROTECTED_CLASS_CHOICES`. We would then remove the form order from National origin (including ancestry, ethnicity, and language) and add the correct order to the new elements. We would then make sure that the "Other"javascript was pointed at the correct element.

crt_portal/cts_forms/templates/forms/report_details.html
LindsayYoung: It's not a new alert, it was already there but not styled the same as the other error fields.

crt_portal/cts_forms/tests.py
LindsayYoung: The invalid test case is below: 
https://github.com/usdoj-crt/crt-portal/pull/91/files#diff-18509124e855af21de52f53726ce2005R66

LindsayYoung:
@alexsoble and @ilodidoj the requested changes and explanations are in

crt_portal/cts_forms/model_variables.py
alexsoble: Ooooooh, thanks, this is a great explanation! Can this be a code comment?

LindsayYoung: Ok, that is a good idea. 

I added a slightly longer comment in our `/docs` and put a bread crumb next to the protected class model, and the variable definitions

LindsayYoung:
@alexsoble added documentation and resolved merge conflict. Feel free to merge if you are happy with it

crt_portal/cts_forms/tests.py
alexsoble: Ah, I see. I think I was confused by the difference between the `Valid_Form_Tests` and `Validation_Form_Tests` classes. Thanks!

crt_portal/cts_forms/templates/forms/report_details.html
alexsoble: I see, thanks! So essentially this alert is for back-end errors and the alert below is for front-end errors. In this case the back-end errors should never be triggered because the field is optional and we're trimming down word count to be below the database constraint ... but if somehow a back-end error does appear, it'll be styled right! Appreciate the explanation.

PR #93, opened 2019-10-10
Fix double-label bug in grouped questions template 

alexsoble:
## Before
double-label

## After

single-label

### DIFF
--- a/crt_portal/cts_forms/templates/forms/report_grouped_questions.html
+++ b/crt_portal/cts_forms/templates/forms/report_grouped_questions.html
@@ -23,7 +23,9 @@
       {% endif %}
 
       {% for field in question_group %}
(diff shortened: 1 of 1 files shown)


crt_portal/cts_forms/templates/forms/report_grouped_questions.html
```diff
@@ -23,7 +23,9 @@
       {% endif %}
 
       {% for field in question_group %}
-        <label class="margin-bottom-0">{{ field.label_tag }}</label>
+        <label for="{{ field.id_for_label }}" class="margin-bottom-0">
+          {{ field.label }}
```
alexsoble: `label` vs `label_tag` distinction here: https://docs.djangoproject.com/en/2.2/topics/forms/#looping-over-the-form-s-fields.

PR #103, opened 2019-10-17
View protected class

LindsayYoung:
[View 'Protected Classes' on View all Complaints table.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/130)

## What does this change?
adds protected class

Screen Shot 2019-10-21 at 5 07 20 PM

## Checklist:
check against ACs
### Author

### Reviewer

## Notes for reviewer:
Added a local pally test, you can run that if you set up the user name and password locally. I can add pally set up and tear down steps for login and dummy data to the circle run later, in a seperate ticket.


LindsayYoung:
ready for review when you are back @alexsoble

.pa11y-local.json
```diff
@@ -26,6 +26,17 @@
             "click element #submit-next"
           ],
           "screenCapture": "pa11y-screenshots/page3.png"
+        },
+        {
+          "url": "http://0.0.0.0:8000/form/view",
+          "actions": [
+            "wait for url to be http://0.0.0.0:8000/accounts/login/?next=/form/view",
+            "set field #id_username to pa11y_tester",
+            "set field #id_password to imposing40Karl5monomial",
```
alexsoble: This testing password value would be a good candidate to extract to `.env`.

crt_portal/cts_forms/model_variables.py
```diff
@@ -19,14 +20,30 @@
     'Immigration/citizenship status (choosing this does not share your status)',
     'Religion',
     'Sex or gender identity (including gender stereotypes) or pregnancy',
-    'Sexual orientation',
     'Family, marriage, or parental status',
+    'Sexual orientation',
```
alexsoble: I'm still seeing the old ordering here, what steps should I run to see the new ordering locally?

crt_portal/cts_forms/model_variables.py
```diff
@@ -19,14 +20,30 @@
...
     'Military status',
     'Age',
     'Genetic information',
     'Other',
 )
 
+# used in internal CRT view display
+PROTECTED_CLASS_CODES = {
```
alexsoble: Maybe the variable name could reflect that this is meant for the internal view?

crt_portal/cts_forms/templates/forms/index.html
```diff
@@ -21,44 +23,49 @@ <h1>
...
+    </a>
     </a>
     <table class="usa-table crt-table">
       <thead>
         <tr>
-          <th>Submitted</th>
-          <th>Contact Name</th>
-          <th>Contact Details</th>
-          <th>Violation Summary</th>
+          <th scope="col">Submitted</th>
```
alexsoble: 👍

crt_portal/cts_forms/views.py
```diff
@@ -13,7 +14,17 @@
 @login_required
 def IndexView(request):
     latest_reports = Report.objects.order_by('-create_date')
-    return render_to_response('forms/index.html', {'data_dict': latest_reports})
+    data = []
+    # formatting protected class
+    for l in latest_reports:
```
alexsoble: Small thing, but opportunity to use a semantic variable name instead of single letters here.

LindsayYoung:
@alexsoble I think this should be good to go. I noticed that we should set up some tests for that view, but we can include that in the next PR that touches that table.

LindsayYoung:
Ok, the default state is much better now @alexsoble

crt_portal/cts_forms/views.py
```diff
@@ -18,11 +18,18 @@ def IndexView(request):
     # formatting protected class
     for report in latest_reports:
         reports = []
-        for p_class in report.protected_class.all().order_by('-form_order'):
+        for p_class in report.protected_class.all().order_by('form_order'):
             if p_class.protected_class is not None:
```
alexsoble: Having to call `p_class.protected_class` makes the code difficult to read; it's hard to know how or why the second `protected_class` is different than the first `p_class`.

Looking at the `ProtectedClass` model, I'm seeing that the `protected_class` field is a character field. I'm thinking this is really the name of the protected class, not the protected class of the protected class — does that sound right to you, too?

What do you think about a refactor to change the field to `name`? Since we're in the early stages of building out the app, a refactor for clearer field names could improve the code in a big way going forward.

crt_portal/cts_forms/views.py
```diff
@@ -18,11 +18,18 @@ def IndexView(request):
...
-                reports.append(PROTECTED_CLASS_CODES.get(p_class.protected_class, p_class.protected_class))
+                code = PROTECTED_CLASS_CODES.get(p_class.protected_class, p_class.protected_class)
+                if code is not None:
```
alexsoble: Code should never be `None`, because line 23 specifies that if no value for is found in the `PROTECTED_CLASS_CODES`, the longer protected class name will be supplied as a default. Guarding against null values is good, but two guards back-to-back seems like too many. Also, we never handle the `if code is None` branch — I feel like if we are going to check for `code` being `None`, we should handle the possibility that `code is None`.

crt_portal/cts_forms/views.py
```diff
@@ -18,11 +18,18 @@ def IndexView(request):
...
+                    reports.append(code)
+                elif report.other_class is None:
```
alexsoble: I don't understand why we check for `report.other_class` being `None` only if `code is not None` — the two seem logically independent to me.

crt_portal/cts_forms/views.py
alexsoble: We talked this through and decided that renaming the field to `protected_class_name` could be a good move for a follow-on PR.

crt_portal/cts_forms/views.py
alexsoble: We refactored this and added a code comment to explain!

PR #110, opened 2019-10-23
[WIP] Compress on deploy

LindsayYoung:
JS is not working in dev, this is meant to fix it


LindsayYoung:
This was less straight forward than I thought it would be. I think we can remove the file compression and  add this as a task later that we can prioritize.

PR #114, opened 2019-10-24
Protected Class front-end changes

alexsoble:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-form-design-flows-5d7a67fa86a8cd0001581ea0/issues/18f/crt-django/60)

## What does this change?

+ Add blue highlight to the "Other" field as per the design spec.
+ Update spacing to match the design spec. 
+ Fix small bug with leading whitespace on checkbox labels.
+ Update HTML so that all `<div>` and `<label>` tags are closed. 
+ Simplify logic in HTML template for `report_class.html`. 
+ Uncomment/restore connecting line in progress bar wizard. 

## What issues does this NOT address? 

+ A11Y issues and browser compatibility issues: 
  + We learned from our review session with Terri Youngblood that there are multiple a11y issues to address, such as the error message not announcing properly and a more descriptive label required for "Please describe 'other'." 
  + The review session also surfaced IE compatibility issues, such as the word count function not working in Terri's IE browser.  
+ Alert state visual placement / design requires further conversation.

## Screenshots:

page2

### "Other" checked state

other-checked-state


alexsoble:
Synced up on this work with @LindsayYoung in a 1:1 — we'd like to ship these front-end improvements and get further input from the designers on a few topics.

☝️ Failing pa11y tests are still mysterious to us and require some further investigation:

```
 > http://127.0.0.1:8000/report - Failed to run

Errors in http://127.0.0.1:8000/report:

 • Error: Pa11y timed out (30000ms)
```

LindsayYoung:
The Pa11y test was failing because the `fieldset` attribute was missing a `legend`.

Looking at it locally, if there are errors, I run just the troublesome page. It looks like we were trying to use aria label and instead of fieldset, but pa11y doesn't like that.

```
Errors in http://0.0.0.0:8000/report:

 • Fieldset does not contain a legend element. All fieldsets should contain a legend element that describes a description of the field group.

   (#report-form > fieldset)

   <fieldset class="usa-fieldset" aria-label="Protected Class"> <label for="" cl...</fieldset>
```

crt_portal/cts_forms/templates/forms/report_base.html
```diff
@@ -34,9 +34,7 @@ <h1 class="padding-top-5 margin-0">
             </li>
           {% endfor %}
         </ol>
-        <!-- uncomment when there is more than one step on form
-          <div id="connecting-line"></div>
-        -->
+        <div id="connecting-line"></div>
```
LindsayYoung: 🥇

crt_portal/cts_forms/templates/forms/report_class.html
```diff
@@ -1,39 +1,41 @@
 {% extends "forms/report_base.html" %}
 {% load static %}
 {% block form_questions %}
+
 <fieldset class="usa-fieldset" aria-label="{{ current_step_name }}">
-  <legend class="usa-sr-only">{{ current_step_name }}</legend hidden>
```
LindsayYoung: Mind if we add the legend back?  I think we could remove the aria label. That way, I think it will be accessible and tests will pass.

alexsoble: Oooooh, great catch! Thanks for spotting this @LindsayYoung! 🏆 

A couple of notes here on the `<legend>` implementation that I removed, for learning purposes: 

1. The `hidden` attribute is on the closing tag, should be on the opening tag.
2. The `hidden` attribute and `usa-sr-only` class shouldn't be on the same element. The `hidden` attribute will hide an element for both visual users and screen reader users, so that conflicts with the intent of `usa-sr-only`.

If the element were really `hidden`, my expectation is that pa11y would return the same failure as if the element were not there at all. So maybe this implementation was working because the `hidden` attribute on the closing tag was ignored.

Will investigate further this afternoon after a few meetings wrap up!

alexsoble:
Oooooh, we have a much better error message here now thanks to #116:

```Errors in http://127.0.0.1:8000/report:

 • Fieldset does not contain a legend element. All fieldsets should contain a
   legend element that describes a description of the field group.

   (#report-form > fieldset)

   <fieldset class="usa-fieldset" aria-label="Protected Class"> <label for=""
   cl...</fieldset>
```

Thank you @LindsayYoung!

crt_portal/cts_forms/templates/forms/report_class.html
alexsoble: In this case, the text that's currently marked up as the label for protected class describes the different checkbox input buttons:

> "Do you believe you were treated this way because of any of the following characteristics or statuses that apply to you?"

Semantically, that feels equivalent to a `<legend>`. I switched that text over to `<legend>` which addresses the pa11y issue, has no style impact, and feels like the right thing semantically. For more context: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

alexsoble:
CircleCI is hanging and the build is just stuck on "queued." Ugh.

LindsayYoung:
It passed 🎉

alexsoble:
I like the idea of merging it now and handling alert changes + style tweaks
on a separate PR!

On Oct 28, 2019, at 5:30 PM, Lindsay Young <notifications@github.com> wrote:

﻿

*@LindsayYoung* approved this pull request.

This is looking good! Thanks for the a11y fix! This looks good to me. This
can be merged right now, but I didn't know if you wanted it merged or if
you wanted to use it for the alert state fix.

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/usdoj-crt/crt-portal/pull/114?email_source=notifications&email_token=AAYPSHJ3IR3YHWQ5GWKQF5LQQ5RZPA5CNFSM4JEV75H2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCJPIYWQ#pullrequestreview-308186202>,
or unsubscribe

PR #116, opened 2019-10-25
Pa11y run individually

LindsayYoung:
This should give us less flaky tests and better error messages in circle ci


alexsoble:
Oh this is an awesome fix @LindsayYoung, I'm really glad you thought of this!

As a review step, I'm wondering if we could create a commit on this branch that introduces an a11y mistake on one of the pages (and then back that change out of course), so that we can see the improved CircleCI error message?

alexsoble:
☝️ Woohoo! Much better error message! 

```Errors in http://127.0.0.1:8000/report:

 • Img element is the only content of the link, but is missing alt text. The
   alt text should describe the purpose of the link.

   (#report-form > a)

   <a href="http://example.com/"><img src="example.jpg" alt=""></a>
```

alexsoble:
Hmmm, reverting the latest commit doesn't seem to kick off a new CircleCI build. But 8d6aafe963c68a33af2fe6bbe55e3dc13be8583f was green — so this looks good to go!

PR #123, opened 2019-10-30
View test

LindsayYoung:
Fixes CRT view bug and adds tests to make sure it stays fixed.

### Author

+ [x] Check for [accessibility](/docs/a11y_plan.md).
+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

+ [ ] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.)
+ [ ] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

## Notes for reviewer:

See [PR instructions doc](https://github.com/usdoj/crt-portal/blob/master/docs/pull_requests.md) for full pull request review instructions.

PR #126, opened 2019-10-30
Refactor to use named keys for data in View Table

alexsoble:
## What does this change?

+ Refactor for more semantic code (named dictionary keys instead of positional array lookups). See discussion here: https://github.com/usdoj-crt/crt-portal/pull/123#issuecomment-548021671.


LindsayYoung:
Nice, this is better than it was 5 years ago! lol

PR #131, opened 2019-11-04
Set alert role using JS instead of rendering it server-side

alexsoble:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/167)

## What does this change?

Make the alerts read properly on a screen reader.

Learning that JavaScript is better for adding `role="alert"` than server-side rendering. From Mozilla:

> The most important thing to know about the alert role is that it is for dynamic content. It is perfect for situations such as when a user fills out a form and JavaScript is used to add an error message - the alert would immediately read out the message. It should not be used on HTML where the user hasn't interacted with it. For example, if a page loads with multiple visible alerts scattered throughout, none would be read because they are not dynamically triggered.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role
### Author

### Reviewer


LindsayYoung:
Works on voiceover!
Screen Shot 2019-11-04 at 6 14 11 PM

PR #136, opened 2019-11-08
De-conflicting 2 migrations labeled 13

LindsayYoung:
Debugs the issue where we had conflicting migrations https://app.circleci.com/jobs/github/usdoj-crt/crt-portal/484/parallel-runs/0/steps/0-103

It works for me locally and tests pass in circle. Try it locally, it might give you a warning, or it might get confused if your data base already applied them in a different order.


alexsoble:
The migration applies cleanly for me locally and the form works as expected. 👍

PR #143, opened 2019-11-14
Remove generated files

LindsayYoung:
- These shouldn't be in source control since they are generated. 
- It was causing unnecessary merge conflicts

Also merged in Master to make sure we were not missing anything upstream. 

Works on circle and cloud.gov:
https://app.circleci.com/jobs/github/usdoj-crt/crt-portal/556

Easiest way to test locally is to delete the files, watch the files appear when you run `docker-compose up`. Make sure you are not running `gulp watch` at the same time.


package.json
```diff
@@ -35,7 +35,7 @@
     "pa11y": "^5.2.0",
     "pa11y-ci": "^2.3.0",
     "path": "^0.12.7",
-    "postcss-cli": "^4.1.0",
```
LindsayYoung: This was causing the build to fail, so I upgraded it and did a quick check. It looks to be in order but if you could double check, that would be great.

LindsayYoung: This wasn't the cause of the failure but I would rather keep the upgrade now.

PR #148, opened 2019-11-15
We need to rebuild the sass to make sure the binaries are correct.

LindsayYoung:
# What does this change?
Always rebuilds node-sass to make sure the binaries are correct. 

https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/174


alexsoble:
Don't see the errors locally anymore! Don't totally understand how or why this works so maybe let's discuss this together during our next 1:1 or pairing time!

PR #149, opened 2019-11-18
Assign ADM

LindsayYoung:
[Assign newly created record to ADM](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/91)

## What does this change?
Adds assigned_section attribute and creates a method that we can expand on for auto generating the section assignment at form submission.

crt-view

## Checklist:
(copied from tickef)
Acceptance criteria
1) Given a complaint has been submitted via
-- online intake form by a Public user
-- manual creation by CRT user
When I view the 'View All Complaints' table
Then I view a **column called 'Routed'**

2) Given a new complaint has been submitted via
-- online intake form by a Public user
-- manual creation by CRT user
When I view the 'View All Complaints' table
Then I see

 the **section abbreviation as ADM in the 'Routed' column**
3) given that I'm using a screen reader, when I get to a cell in this column it should announce "I-E-R, Routed" ([cell content, column name]).

I expect this to work as intended already based on the other tables. Preference for keeping it consistent with what sighted people would see vs having it read the section name
### Author

### Reviewer

## Notes for reviewer:
My voice over results were not what I expected, but I am not convinced I was navigating the page the right way. It the column is scoped as "col" so my understanding is that that should add the right labels for VO.


LindsayYoung:
Thanks for the feedback, @alexsoble!!

PR #150, opened 2019-11-18
Organize Sass files

alexsoble:
## What does this change?

Organizes our Sass code into smaller files for better maintainability, readability, and developer experience. 

Goal here is a small, pure refactor PR — keep all front-end styling the same while improving the structure behind the scenes.

Further improvements to come; added TODO notes about removing targeting of styles against raw HTML elements. 

### Form Page 1

page1

### Author

+ [x] Checked to make sure no regressions in front-end styles.

### Reviewer

+ [ ] Checked to make sure no regressions in front-end styles.


LindsayYoung:
This is missing the status dependent styling on the CRT table: https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/sass/_uswds-theme-custom-styles.scss#L262

alexsoble:
@LindsayYoung Ah, great catch ... because I hadn't pulled down the latest from `develop` when I started pulling styles out of the mega-file ... thanks for catching!

crt_portal/static/sass/custom/table.scss
```diff
@@ -14,4 +14,33 @@
...
+  color: $white !important;
+  background-color: $blue-warm-vivid-70;
+  padding: 4px;
+  padding-left: 10px;
+  padding-right: 10px;
+  border-radius: 1.5rem;
+}
+
+.status-open {
+  background-color: $gold-20v;
```
LindsayYoung: We need to add this to the colors as well

     $gold-20v: #FFBE2E;

alexsoble: I noticed that in `colors.scss` we had defined:

```
$gold: #FFBE2E;
```

Since these are the same values I think we can just use a single `$gold` for now.

alexsoble: This compiles correctly for me. If you feel strongly about the variable name, we can get more specific than `$gold` — but elsewhere we're spelling out "vivid" as in `$blue-vivid-40`, so I'd want to keep that consistent.

PR #155, opened 2019-11-18
Adds .nvmrc and updates README

el-mapache:
## What does this change?

* Ensures that we all have the same compiled npm packages
* Adds explicit install instructions to the README
* Adds sass:watch command to our npm scripts in package.json. This ensures
that the `gulp watch` uses the version of gulp specified in our package.json,
not user's system-wide gulp installation (which may or may not even exist)

### Reviewer

+ [x] To test, I believe deleting your `node_modules` folder and running through the install commands in the README will be sufficient to validate that everything is working as it current is.

### Notes to reviewer

We should also add code to install and source `nvm` in the `Dockerfile`. I took a stab at it but ran into issues running the `source` command. I'm not sure what specific bash programs the `Dockerfile` exposes, so troubleshooting this is sliiightly out of my wheelhouse. A followup PR, or guidance on how to best update the `Dockerfile` in this PR would be great!


LindsayYoung:
Thanks for catching this!

I was running 12 locally, but I can see that I set up the docker container with 10. Since it has been working with 12, my inclination would be to do a follow up PR to bump the version in the docker file to 12. If that sounds good, I can merge this and upgrade the builds. 

I was wondering, since we are specifying the version in the install command and we don't need to switch node environments in the container, do we need NVM? Or does NVM do more than I am thinking?

Local docker container with 10:
https://github.com/usdoj-crt/crt-portal/blob/aca2c0d7d5430acaaa5adef0002e34bd3d6380d0/Dockerfile#L23

That is also what we have for production deployments:
https://github.com/usdoj-crt/crt-portal/blob/develop/.circleci/config.yml#L209

For pre-deployment tests, we are pinning Python but not node because we are using the pre-built images form circle with the specified version of Python and the LTS release of node. We could replace that image, we are using that for convenience, since it also has browsers installed, good for pa11y testing. 
https://github.com/usdoj-crt/crt-portal/blob/aca2c0d7d5430acaaa5adef0002e34bd3d6380d0/.circleci/config.yml#L11

el-mapache:
I think a followup PR to bump the node version in docker is a great idea.

We essentially only need NVM to ensure that dependencies are installed and compiled using the same node version across developer machines, and It should also prevent the `package-lock.json` file from getting updated unnecessarily.

PR #158, opened 2019-11-19
Sass improvement: Use explicit class for emphasis text

alexsoble:
## What does this change?

+ Incremental Sass cleanup: use explicit class for emphasis text rather than applying styling to raw `<legend>` HTML elements.
+ This should be a pure refactor PR and should not have any effect on the front-end styling. 

### Author

### Reviewer

## Notes for reviewer:

+ Please check to make sure this doesn't introduce any style regressions ... thank you!

### DIFF
--- a/crt_portal/static/sass/custom/typography.scss
+++ b/crt_portal/static/sass/custom/typography.scss
@@ -17,12 +17,11 @@ h3 {
 }
 
 // TODO (ARS): Don't target raw HTML elements with styles.
-p, em, h2, label:not(.big), span:not(.big) {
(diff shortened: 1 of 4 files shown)


crt_portal/static/sass/custom/typography.scss
```diff
@@ -17,12 +17,11 @@ h3 {
 }
 
 // TODO (ARS): Don't target raw HTML elements with styles.
```
alexsoble: I bet there's a way I can fix this in my next PR by adjusting our USWDS settings rather than using an override.

PR #159, opened 2019-11-19
Sass improvement: Factor out shared styles in table.scss

alexsoble:
## What does this change?

+ When I was factoring our Sass into different files, I noticed a quick opportunity to refactor by pulling shared styles into a single class.
+ This should be a pure refactor PR and should not have any effect on the front-end styling. 

### Author

### Reviewer

## Notes for reviewer:

+ Please check to make sure this doesn't introduce any style regressions ... thank you!

### DIFF
--- a/crt_portal/cts_forms/templates/forms/index.html
+++ b/crt_portal/cts_forms/templates/forms/index.html
@@ -43,7 +43,11 @@
     {% if data_dict %}
       {% for datum in data_dict %}
         <tr class="tr-status-{{ datum.report.status }}">
-          <td><span class="status-{{ datum.report.status }} row-status-{{ datum.report.status }}">{{ datum.report.status }}</span></td>
+          <td>
+            <span class="status-tag status-{{ datum.report.status }} row-status-{{ datum.report.status }}">
+              {{ datum.report.status }}
+            </span>
+          </td>
           <td>{{ datum.report.assigned_section }}</td>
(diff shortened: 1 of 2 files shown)


crt_portal/cts_forms/templates/forms/index.html
```diff
@@ -43,7 +43,11 @@ <h1>
     {% if data_dict %}
       {% for datum in data_dict %}
         <tr class="tr-status-{{ datum.report.status }}">
-          <td><span class="status-{{ datum.report.status }} row-status-{{ datum.report.status }}">{{ datum.report.status }}</span></td>
+          <td>
+            <span class="status-tag status-{{ datum.report.status }} row-status-{{ datum.report.status }}">
```
LindsayYoung: I think you added an extra `row-status-new` class.

alexsoble: The only additional class here is the shared `status-tag`; would you mind taking another look @LindsayYoung?

LindsayYoung: I don't see `row-status-new` defined in the css. I don't think that class was supposed to be in there.

alexsoble: Agreed!

LindsayYoung:
thanks for the extra clean up!

PR #160, opened 2019-11-19
Updates node to 12.x in circle config + Dockerfile

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/180)

## What does this change?

* Follow-up PR to #155, updates Docker version of node to 12.x 

## Notes for reviewer:

I think these are the only two places we need to make this update. Let me know if I missed something!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,7 +94,7 @@ jobs:
       - run:
           name: Node install
           command: |
-            echo "deb https://deb.nodesource.com/node_10.x stretch main" > /etc/apt/sources.list.d/nodesource.list && \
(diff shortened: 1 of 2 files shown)


LindsayYoung:
This works for me, I did need to delete my local node packages and rebuild the container. 

Flagging @alexsoble @dwechsle @ilodidoj :
- If you get `throw new Error(errors.missingBinary());` on start up, delete your node files and rebuild with `docker-compose up`
- We added instructions for adding NVM locally for your macs. This will make front end devloping easier because it makes sure your local `gulp watch` doesn't get out of sync with the project dependencies.

PR #166, opened 2019-11-20
Sass improvement: Use USWDS settings to set default ink color 

alexsoble:
## What does this change?

+ Use USWDS settings to set default ink text color instead of manual overrides which set styles on raw HTML elements.
+ This should be a pure refactor PR with one exception: make the text color on the View All table match the text color in the rest of the app.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/static/sass/custom/table.scss
+++ b/crt_portal/static/sass/custom/table.scss
@@ -26,7 +26,7 @@
 // Styles for complaint statuses in View All Table
 
 .status-new {
-  color: $white !important;
+  color: $white;
   background-color: $blue-warm-vivid-70;
 }
 
--- a/crt_portal/static/sass/custom/table.scss
+++ b/crt_portal/static/sass/custom/table.scss
@@ -1,10 +1,10 @@
 .crt-table {
   @extend .usa-table;
   font-size: 14px;
-  color: $gray-90;
   line-height: 22px;
   background: $white;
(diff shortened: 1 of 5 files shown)


crt_portal/static/sass/custom/table.scss
```diff
@@ -26,7 +26,7 @@
 // Styles for complaint statuses in View All Table
 
 .status-new {
-  color: $white !important;
```
alexsoble: Safely removing an `!important` is a mini code quality win!

crt_portal/static/sass/custom/typography.scss
```diff
@@ -16,11 +16,6 @@ h3 {
   @include u-text('primary-dark')
 }
 
-// TODO (ARS): Don't target raw HTML elements with styles.
-p, em, h2, label:not(.em-text), span:not(.em-text) {
```
alexsoble: Removing styling on raw HTML elements is a mini code quality win!

crt_portal/static/sass/custom/table.scss
```diff
@@ -1,10 +1,10 @@
 .crt-table {
   @extend .usa-table;
   font-size: 14px;
-  color: $gray-90;
```
alexsoble: I believe the text color in the View All table should be `gray-warm-80`, same as the default in rest of the app, as noted here: https://github.com/18F/crt-portal/issues/25.

PR #171, opened 2019-11-22
 i18n JavaScript translations

el-mapache:

## What does this change?

This PR adds a few set-up pieces that should allow us to internationalize strings in our JS code. 

To wit:

* Adds a route to expose the JS helpers in the `JavaScriptCatalog`, and an accompanying script tag in `report_base.html` to load it.

* Small tweaks to the `word_count.js` file for translation testing purposes. Also wraps that code in an anonymous function to avoid polluting the global namespace.

* Adds an explicit `LOCALES_PATH`. I'm unsure if this is needed

## Notes for reviewer:

I rebased off of @alexsoble's PR, so the files changed is a bit noisy here. The only files that I touched for this PR were:

`README.md`
`settings.py`
`urls.py`
`report_class` and `report_base.html`
`word_count.js` 

I've left what I hope to be fairly comprehensive notes in the zenhub issue, but I'll recap briefly here!

1). Having the `locale` folder live inside the `cts_forms` folder loads the JS translations, but running `makemessages` to generate the `.po` files does _not_ work — django can't seem to find the `locale` folder when it isn't co-located with the `static` folder

2) `makemessages` runs on every JS file in the repo, and generates a `.js.py` file for each, so it may be prudent to figure out a way of putting all the JS strings in a single file


el-mapache:
@alexsoble Not sure what we should do with this PR...the JS translating sort of works with various caveats, but I'm not really sure if it's worth merging just to capture some of the set-up process. Particularly when we don't need translations right away! Let me know your thoughts !

el-mapache:
Closing as both of @LindsayYoung 's PRs implementing the results of the i18n spike have been merged! 🎉

PR #176, opened 2019-11-25
Add changepassword command to README

alexsoble:
## What does this change?

Add `changepassword` command to README for reference and ease of copy-paste.


alexsoble:
@el-mapache Yay! Please merge if you don't mind, that way I'm not merging my own work!

PR #179, opened 2019-11-26
Document hard-resetting the app/database with docker prune

alexsoble:
## What does this change?

+ Document how to use `docker system prune --volumes` if a dev needs to rebuild the app and db from a clean slate.


alexsoble:
@el-mapache Great call, added a warning in the commit above. ☝️

PR #180, opened 2019-11-26
Apply text-transform to status badge only in form viewer

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/184)

## What does this change?

* Makes sure only the table headings and the text in the status badges have an `uppercase` text transformed applied.

Screen Shot 2019-11-26 at 12 29 10 PM

### DIFF
--- a/crt_portal/static/sass/custom/table.scss
+++ b/crt_portal/static/sass/custom/table.scss
@@ -18,12 +18,13 @@
   td {
     vertical-align: top;
   }
-}
-
-.tr-status-new {
-  td {
-    font-weight: bold;
-    text-transform: uppercase;
+  
+  tbody {
+    .tr-status-new {
+      td {
+        font-weight: bold;
+      }
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


Jacklynn:
Looks good to me!

crt_portal/static/sass/custom/table.scss
```diff
@@ -20,11 +20,9 @@
   }
 }
 
-.tr-status-new {
```
LindsayYoung: this is for the entire row of new forms, I think they are still supposed to be bolded

crt_portal/static/sass/custom/table.scss
```diff
@@ -20,11 +20,9 @@
...
-  td {
-    font-weight: bold;
-    text-transform: uppercase;
-  }
+.status-new {
```
LindsayYoung: I think the uppercase transform needs to move to `status-tag`  so open and closed are also capitalized

el-mapache:
@LindsayYoung Thanks for taking a look! PR has been updated:

Screen Shot 2019-11-27 at 8 54 35 AM

Now records with a status of `new` will be properly bolded, but maintain their standard casing.

PR #181, opened 2019-11-27
View correspondant information on complaint detail page

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/24)

## What does this change?

* .Adds a new python view for route details
* Adds HTML templates and scss for the complaint page
* Separates the two HTML templates for our custom complaint admin to their own folder
* I left in some debug code that will only show up on our local machines (I think... : p), I'll remove it when I get more comfortable with the data model

#### What's next

* A potential follow-up PR to make the routes more RESTful. Perhaps `complaints` and `complaints/:id` vs. the current `form/view` and `form/:id`, which I think are a bit generic.

#### What is not here!

* Pixel-perfect spacing, potentially slightly incorrect color choices, etc. There aren't annotated mocks for this page yet that I saw; I think this is a good place from which to make further refinements!

Screen Shot 2019-12-02 at 2 48 26 PM

### GIF
table-nav

### Author

### Reviewer

## Notes for reviewer:

I don't super love the hover effect currently on the table. The box shadow effect doesn't _quite_ work because `tr` aren't `display: block`, which collapses the table in various weird ways (here removing the inner borders). There are some tricks to try with `:first` and `:last` child pseudo selectors on the `td` elements, but this requires more work, and I'd rather wait until the next round of updates to the styling of the table.


el-mapache:
👁  Opening for visibility! 👁

LindsayYoung:
Exciting!

alexsoble:
I looked at the code and at the page locally — both look fantastic @el-mapache!

crt_portal/cts_forms/tests.py
```diff
@@ -340,6 +343,17 @@ def test_view_all_unauthenticated(self):
         self.assertEqual(response.status_code, 302)
         self.assertRedirects(response, '/accounts/login/?next=/form/view')
 
+    def test_view_report_details_authenticated(self):
```
alexsoble: 🎉Wooohooo for auth tests! 🎉

crt_portal/static/sass/custom/complaint.scss
```diff
@@ -0,0 +1,67 @@
...
+    text-decoration: none;
+
+    &:visited {
+      color: $blue-50;
+    }
+  }
+}
+
+.complaint-body {
+  transform: translateY(calc(-120% + #{$complaint-header-min-height}));
```
alexsoble: @el-mapache Curious what this line of code is doing — I don't know how to read it.

el-mapache: Yeah this got a little muddled. I have a simpler calculation that can express the offset we want here. I'm using the `transform: translate` syntax because I think it conveys the intention that this element is going to be _placed_ somewhere on the page better than `margin-top`!

el-mapache: new calc is 
```
$complaint-body-offset: calc(
  -#{$twenty-percent-mult} * #{$complaint-header-min-height}
);
```

el-mapache: And if you are asking specifically about the `transform: translateY` call, is basically just means: move the element `n` units (`px`, `%`, etc) vertically along a 2D plane, relative the element's current position in the browser window. 

So, if the element is `300px` from the top of the `window`, and you apply a `translateY(40px)`, the element will move down an additional 40px to 340px offset from the window.

LindsayYoung:
This is really great! 

One small thing. We might want to make sure we are handling nulls consistently here, on the list view we are using em dashes as the fall back when there is no value.

el-mapache:
@LindsayYoung Do you think  we should use `N/A` in the table view as well? The designs for this page all feature `N/A`, so I'd be in favor of moving to moving to that.

crt_portal/cts_forms/templates/forms/complaint_view/show.html
```diff
@@ -0,0 +1,84 @@
...
+                    Edit
+                  </button>
+                </div>
+              </div>
+              <table class="usa-table usa-table--borderless complaint-card-table">
+                <tr>
+                  <th>Name</th>
+                  <td>
+                    <strong>
+                      {{ data.contact_last_name }}, {{ data.contact_first_name }}
```
LindsayYoung: Names are not required for the form so these will need a default too

crt_portal/cts_forms/templates/forms/complaint_view/show.html
```diff
@@ -0,0 +1,84 @@
...
+          <h2 class="text-uppercase">
+            Complaint <span class="text-gold">{{ data.id }}</span>
+          </h2>
+          <p>
+            <span class="text-gold text-uppercase">Received:  </span>
+            <span class="text-white">{{ data.create_date|date:'g:s a F j, Y' }}</span>
+          </p>
+        </div>
+        <div class="tablet:grid-col-3 tablet:grid-offset-1 text-right">
+          <a class="usa-button usa-button--light close-record" href="/form/view">
```
LindsayYoung: Not sure if we want to make this part of this PR but I wanted to note that when we add sort and filter options, we will want to pass that to this template probably something like adding next={{ page_format.page_args }}&page={{ page_format.page }} to the url and then adding that to the close record link.

crt_portal/cts_forms/views.py
```diff
@@ -50,11 +53,32 @@ def IndexView(request):
...
+        'data_dict': data,
+        'page_format': page_format,
+        'page_args': page_args,
+        'sort_state': sort
+    })
+
+
+@login_required
+def ShowView(request, id):
+    report = Report.objects.get(id=id)
```
LindsayYoung: I think we might want to use get_object_or_404() here.

crt_portal/static/sass/custom/complaint.scss
alexsoble: @el-mapache I understand much better now! I like the distinction in intention you drew between `translate` and `margin`. Thanks!

crt_portal/cts_forms/templates/forms/complaint_view/show.html
el-mapache: Are both of these values present in the codebase now? Maybe we can go over the sorting changes that got merged yesterday? I'd love to get it into this PR and some assistance would be great 😄

LindsayYoung: Happy to pitch in! would you like me to commit to this PR or make a PR targeting your PR?

el-mapache: If it's not too much trouble, a commit here is great!

crt_portal/cts_forms/views.py
el-mapache: I added a test for this, just wanted to note that the `test_required_user_logging` seems slightly brittle as written; it needs to be manually updated each time a new suite of tests is written involving an authenticated user. Perhaps there is another metric not tied to a specific number of database operations we could use for this test? I'm not familiar enough with the ins and outs of testing django apps quite yet, so I don't have any specific proposals, but I did want to flag it here. Happy to make an issue or document this elsewhere if it seems important!

crt_portal/cts_forms/templates/forms/complaint_view/show.html
LindsayYoung: added!

crt_portal/cts_forms/views.py
LindsayYoung: Generally, I create an object in the set up for the class, you can save the example record and then save `self.test_report_id = Report.pk` so you can call  reverse look up on `f'form/{self.test_report_id}'` in your tests and you know it exists.

el-mapache: Maybe we could talk about this a bit more tomorrow, I don't think I'm explaining what I'm seeing correctly 😬

el-mapache:
@LindsayYoung Is this safe to merge at this point?

LindsayYoung:
Thanks for the wait, I wanted to do the staging release first, that will give plenty of time for review.

PR #190, opened 2019-12-04
Add per-instance FORM_AUTOCOMPLETE_OFF flag 

alexsoble:
## What does this change?

+ Add an environment variable flag for turning form autocomplete off for an instance of the app; default to keeping form autocomplete on if no flag is set.
+ We want to disable form autocomplete on staging only; adding a flag rather than hard-coding the value will let us adjust per-instance easily. 

## Notes for reviewer:

+ To test, set `FORM_AUTOCOMPLETE_OFF=True` (or any other string value) in your `.env` file, reload the site, and check form autocomplete.


LindsayYoung:
I think you will want to add the auto complete off variable here: 
https://github.com/usdoj-crt/crt-portal/blob/develop/manifest_staging.yaml#L8

alexsoble:
@LindsayYoung Ooooh, thank you! I had a Heroku-style mental model of adding ENV variables via the command line / UI, thanks for pointing me to the YAML — will add that now!

alexsoble:
@LindsayYoung Added in 3bee9001bf944d7019724205782e4811318e9a93, thanks again for pointing that out!

alexsoble:
@LindsayYoung I think this should be good to 🚢!

LindsayYoung:
This does add the autocomplete toggle and renders correctly, but chrome does autofill anyway. It seems to be a Chrome issue... still looking into it.

alexsoble:
@LindsayYoung Even when you set the flag in your `.env` file?

Here's what I see locally on `develop`, with `FORM_AUTOCOMPLETE_OFF=True` set in `.env`:

Screen Shot 2019-12-05 at 12 24 08 PM

Here's what I see locally on this branch, with `FORM_AUTOCOMPLETE_OFF=True` set in `.env`:

Screen Shot 2019-12-05 at 12 25 20 PM

Do you see the same thing?

PR #196, opened 2019-12-06
Adds private bucket

LindsayYoung:
Refactored s3 settings to account for having multiple buckets and added the details for the private bucket we are using for SSO


LindsayYoung:
I am going to keep everything in the sso pr

PR #200, opened 2019-12-09
Add ReportFilter class and specs

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/37)

## What does this change?

This PR adds preliminary filtering support for the `Report` model. I wanted to split this issue up into at least two PRs because:
  * the back-end can be tested in isolation
  * the front-end is a pretty significant feature lift!

Let me know if this code seems weird or unidiomatic, I'm still trying to wrap my head around the 'python way'!

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/views.py
+++ b/crt_portal/cts_forms/views.py
@@ -14,13 +14,15 @@ from django.conf import settings
 from .models import Report, ProtectedClass
 from .model_variables import PROTECTED_CLASS_CODES
 from .page_through import pagination
-
+from .filters import ReportFilter
 
 SORT_DESC_CHAR = '-'
 
 
 @login_required
 def IndexView(request):
+    report_filter = ReportFilter(Report, request.GET.dict())
+
     # Sort data based on request from params, default to `created_date` of complaint
     sort = request.GET.getlist('sort', ['-create_date'])
     per_page = request.GET.get('per_page', 15)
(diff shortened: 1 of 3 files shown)


crt_portal/cts_forms/views.py
```diff
@@ -14,13 +14,15 @@
...
 from .page_through import pagination
-
+from .filters import ReportFilter
 
 SORT_DESC_CHAR = '-'
 
 
 @login_required
 def IndexView(request):
+    report_filter = ReportFilter(Report, request.GET.dict())
```
LindsayYoung: So one thing that we may want to plan ahead on - `request.GET.dict()` will only return the last peram if it is a repeated peram. So if we want to search for two sections or extend the code to search for two or more things, in the future it won't work.

If we use something like `request.GET.getlist('assigned_section')` that will give us a list for the filter. (We could probably incorporate that part of the filter code in a more elegant way.)

I'll defer to you if you want to tackle that now, but I am fine merging as is if you don't want to worry about that yet.

el-mapache: Oh that is a great point, I guess I was sort of hoping django would group everything into a bucket for you automatically 😄 . I'll just make that change now.

el-mapache:
Adjusting the filter class actually requires a bit more work than anticipated. I'm merging this now so I can continue with the front-end, and passing off the last bit to @LindsayYoung . Thanks 😄

PR #205, opened 2019-12-12
implements backend filter functionality for matching fields

LindsayYoung:
Creates and filter function we can extend later, it provides filters when people want a field to provide one or more matches.

Need to check in with @el-mapache about what `filters` should pass to the front end 
Need to add more tests

### DIFF
--- a/crt_portal/cts_forms/tests.py
+++ b/crt_portal/cts_forms/tests.py
@@ -270,24 +270,22 @@ class CRT_FILTER_Tests(TestCase):
         self.user = User.objects.create_user('DELETE_USER', 'george@thebeatles.com', self.test_pass)
         self.client.login(username='DELETE_USER', password=self.test_pass)
         url_base = reverse('crt_forms:crt-forms-index')
-        url = f'{url_base}?assigned_section=ADM&status=great'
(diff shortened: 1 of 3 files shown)


crt_portal/cts_forms/tests.py
```diff
@@ -270,24 +270,22 @@ def setUp(self):
...
+    #     self.assertFalse(context['filters'].get('state'))
 
     def test_filter(self):
-        expected_reports = filter(
-            lambda x: x.assigned_section == 'ADM',
-            Report.objects.all()
-        )
+        expected_reports = Report.objects.filter(assigned_section='ADM')
         actual_reports = self.response.context['data_dict']
+        print("\n\nACTUAL-----", actual_reports)
```
el-mapache: I think we'll want to remove these print statements.

el-mapache:
@LindsayYoung The only filter we need right now is the assigned section functionality, but what you have here is fine.

LindsayYoung:
@el-mapache since you approved, I added an extra test, and removed all print statements, I am going to merge it now

el-mapache:
Great! I'll pull in to my branch in a bit, hopefully it should integrate smoothly.

PR #206, opened 2019-12-12
Front-end interface to complaint filtering

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/37)

## What does this change?

This PR adds the front-end interface to the filtering work merged in (this PR)[https://github.com/usdoj-crt/crt-portal/pull/200] and (this one)[https://github.com/usdoj-crt/crt-portal/pull/205]. Thanks @LindsayYoung for the awesome assist!

The styles don't quite match the mocks yet; there is some work to be done around styling the table and the filter section on smaller screen sizes. Currently the table is too wide to accommodate both the filters and the table on a laptop.

**Additional changes**:
* `CRTDropdown` widget is now generic and not hardcoded to have a `list` property of `states`
* Adds filters to `sortable-table_heading` inclusion tag so that sorting the table will preserve applied filters.
* Adds JS to control behavior combining filtering and sorting.

### Filters section
Screen Shot 2019-12-12 at 3 37 03 PM

### End of results
Screen Shot 2019-12-16 at 8 18 25 AM

### No results found
Screen Shot 2019-12-16 at 8 19 50 AM

### Widescreen monitors
Screen Shot 2019-12-16 at 7 47 40 PM

### Tablet / laptop view
Screen Shot 2019-12-16 at 7 47 52 PM

### Author

### Reviewer

### TODOS

- [x] Add 'You've reached the end of results' message
- [x] Add 'No records match your search' message
- [x] Ensure filters and sorts populate the URL query string correctly
- [x] Add UI behavior to filter close buttons


el-mapache:
@LindsayYoung i solved the filtering and sorting url param conflict using both JS and django. I think the functionality is working correctly but I'd love to hear (read??) your thoughts!

crt_portal/cts_forms/views.py
```diff
@@ -26,7 +27,7 @@ def IndexView(request):
 
     # Sort data based on request from params, default to `created_date` of complaint
     sort = request.GET.getlist('sort', ['-create_date'])
-    per_page = request.GET.get('per_page', 15)
+    per_page = request.GET.get('per_page', 1)
```
LindsayYoung: The default should be 15

LindsayYoung: Checked locally and that will fix all the tests. I assume you would want me to make the change. I will confirm with you before merging.

LindsayYoung:
This looks like a great start!

I fixed the tests by restoring the per_page default.

From the ACs, we able to select more than one section at a time. We can do that if we go directly to the url `/form/view?per_page=15&sort=-create_date&page=1&assigned_section=VOT&assigned_section=ADM` but I can't seem to be able to construct that in the interface.

Found a small bug where if you go to a detail page and then close the record, when you come back to the view all table, you see the section applied twice.
Screen Shot 2019-12-16 at 10 16 54 AM

I don't see the end of records note. I also added a comment about AC 7 in the card. We might want to simplify that. and use the same logic we use to gray out the right arrow when you are on the last page.

crt_portal/cts_forms/views.py
el-mapache: Oh I can just change it back, I just needed it set to one to check the 'end of results' message logic!

el-mapache:
@LindsayYoung Unfortunately, we can't select multiple filters using the `datalist` element; we'll need to find an external JS library to handle the typeahead select functionality for this use case. I'm going to check in with design and see if they mind having only a single selectable section, with the typeahead functionality coming in a later PR.

LindsayYoung:
I found an error with taking off the filter tags:
If you go to http://0.0.0.0:8000/form/view?sort=location_city_town&sort=location_state&assigned_section=ADM&assigned_section=IER&assigned_section=VOT

Then click the x to remove ADM, it removes Voting. If you go back to that url and click IER, it also removes voting.

It may be removing the last item form the filter list instead of the matching item?

PR #207, opened 2019-12-13
Add text and date filters

LindsayYoung:
## What does this change?
Adds the filters we will need:
- Uses the postgres `to_tsvector` search for some pretty good out of the box text search
- Adds create date range search, (can change date format in the URL if needed later)
- Adds tests

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/tests.py
+++ b/crt_portal/cts_forms/tests.py
@@ -1,5 +1,6 @@
 import secrets
 from datetime import date, timedelta
+import copy
 
 from testfixtures import LogCapture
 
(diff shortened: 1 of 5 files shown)


crt_portal/cts_forms/tests.py
```diff
@@ -442,14 +487,14 @@ def test_required_user_logging(self):
 
             self.assertEqual(
                 cm.check_present(
-                    ('cts_forms.signals', 'INFO', 'ADMIN ACTION by: CLI CLI @ CLI User saved: 5 permissions: <QuerySet []> staff: False superuser: False active: True')
+                    ('cts_forms.signals', 'INFO', 'ADMIN ACTION by: CLI CLI @ CLI User saved: 7 permissions: <QuerySet []> staff: False superuser: False active: True')
```
el-mapache: This is the line I was talking about. I think we should have some mechanism to not have to manually update these permission checks as we add new tests.

LindsayYoung: Yea, good idea- I added some formatting to get the user id dynamically.

PR #209, opened 2019-12-17
Conditional logic - Voting

LindsayYoung:
[Conditional logic - Voting](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/182)

Need to work on template and progress bar.

## What does this change?
If someone ticks the voting primary complaint, the follow up question on election type will appear.

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/views.py
```diff
@@ -126,6 +128,22 @@ def ShowView(request, id):
...
 
+def show_election_form_condition(wizard):
+    # try to get the cleaned data of step 1
+    cleaned_data = wizard.get_cleaned_data_for_step('1') or {'primary_complaint': 'not yet completed'}
+    if cleaned_data['primary_complaint'] == 'voting':
+        return True
+    return False
+
+
+def show_location_form_condition(wizard):
```
el-mapache: I was under the impression that the default `location` question set was always displayed during the 'where' step. If that is the case, do we need to check the negative condition here?

Based on the mocks I see both appearing, so it feels like just the check to explicitly show the `election` section of the form is needed.

LindsayYoung: For Django form wizard, I think of the forms as pages and the fields as questions.

We wanted the election information to be on the location pages, so the approach is to have two versions of the location page. Based on your primary issue response, you will either will get the Location form page which only has the location questions or the ElectionLocation form page, which has the location questions and the election question. 
``` 
                        | (voting)  -->  election location --|
contact -> primary issue|                                    |--> protected class --> details
                        | (not voting) -->  location       --|    
```

el-mapache: Absolutely, I totally get that they are distinct pages. I guess I was just hoping that we could eliminate one of the choices in the dict in `urls.py`, e.g. just have a condition for `2`, since 3 would be the default in all other cases. Seems that is not the case, or might not be useful even if it is. Thanks!

PR #215, opened 2019-12-18
Remove tab indicies from non-actionable content

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/213)

## What does this change?

This PR removes the individual tab stops added to each label and legend. We were advised by our a11ies to only have screen reader stops on actionable content.

To provide users who rely on screen readers with extra contextual information about form fields or question groups, we are leveraging the `aria-describedby` html attribute. Additional field level help text will be moved into the input's `label` element as additional, screen reader only content. 

### Author

### Reviewer

## Notes for reviewer:

You may see a couple of things in this PR that seem weird! And they might be! But i'll attempt to provide a cogent rationale for them anyway!

* In the file `crt_portal/cts_forms/templates/forms/widgets/usa_checkbox_option.html` you'll notice that the `label` elements have `role=label`. While this may seem redundant, I discovered that setting this stops VoiceOver from reading the label twice.

  According to [this issue](https://bugs.webkit.org/show_bug.cgi?id=191743) it seems that `label` elements are converted to text, _and_ the text within them is reported to the screen reader as a child element of the label, with its own text. Setting `role=label` explicitly tells Webkit to correctly report the content within the element once.

* I currently can't get the `city town` or `state` fields to properly read the question group title and italicized help text. This is happen because they don't exist in the same question group. Currently there is no way to add this functionality to the current `QuestionGroup` class. I'm not sure if we should be writing new abstractions or if there is some other obvious thing I'm missing. We could certainly hardcode the strings into the templates for now, as a last resort.

* The same issue occurs in the `other_class` field, since it is a). distinct  from all the other checkbox choices, and it b). has a custom template that doesn't conform to any of our other inputs. Again, maybe we need to hardcode for now?


LindsayYoung:
There is a pa11y check failing on this @el-mapache. I go to the next thing waiting on review and circle back on this one.

el-mapache:
@LindsayYoung Fixed! Thanks for the heads up 😄

PR #217, opened 2019-12-20
Refactor 'other_show_hide' js logic

el-mapache:
## What does this change?

Here, I present a small refactor of the 'other_show_hide' JavaScript logic. The previous file had some non-standard / unnecessary code, as well as duplication of DOM queries. This PR also renames variables to be more explicit about which DOM element they represent.

Finally, I added a few comments that will hopefully provide additional clarity as to why this code is necessary!

### DIFF
--- a/crt_portal/static/js/other_show_hide.js
+++ b/crt_portal/static/js/other_show_hide.js
@@ -1,22 +1,26 @@
-var otherbox = document.getElementById('other-class-option');
-otherbox.setAttribute('hidden', '');
+(function(dom) {
+  /**
+   * The index of the 'other' option on the protected class form.
+   * We target it explicitly becuase while it is currently
+   * the last element in the list of options, it might not always be
+   */
+  var OTHER_OPTION_INDEX = 13;
 
-var OtherIndex = 13;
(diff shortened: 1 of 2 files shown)


crt_portal/static/js/other_show_hide.js
```diff
@@ -1,22 +1,26 @@
-var otherbox = document.getElementById('other-class-option');
-otherbox.setAttribute('hidden', '');
+(function(dom) {
+  /**
+   * The index of the 'other' option on the protected class form.
+   * We target it explicitly becuase while it is currently
+   * the last element in the list of options, it might not always be
```
LindsayYoung: Do we want to warn people to keep the other option as last?

If we don't have other as the last option, we will have to reimplement how we add the other option text and I think it would be hard to have a good no javascript fall back for it. If it is always at the end, it is easy for the non-collapsed version to be generated.

el-mapache: So, you think we should amend this comment to specifically say that the `other` option should always be last? That warning might be a good idea, but I don't think it should live in the JS file. It think it could be co-located with python code?

LindsayYoung: sounds good I can put in a small PR for that.

PR #218, opened 2019-12-20
Fix HTML validation errors

el-mapache:
## What does this change?
Quick little PR to fix some HTML validation issues!

* Removes an extra `div` from the usa-banner
* Moves usa-banner into its own set of partials for easier readability
* Move `base.html` into top level templates folder to improve organization
* Removes unneeded `contentrole` attr from `footer
* Removes unneeded `label` attr from next button `input` node
* Moves `script` tags into the body

### DIFF
--- a/crt_portal/cts_forms/templates/base.html
+++ b/crt_portal/cts_forms/templates/base.html
@@ -1,0 +1,59 @@
+{% load static %}
+<!DOCTYPE html>
+<html lang="en">
+ <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    {# Asking search engines not to index our site while we're in development mode. #}
+    <meta name="robots" content="noindex">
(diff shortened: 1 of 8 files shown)


LindsayYoung:
I think keeping the form base separate from the base will be helpful as we add different types of pages soon. It's not a dealbreaker, just wanted to note it

el-mapache:
@LindsayYoung I'm not sure I follow, I think there is still a distinction between the pages with `base` and `report_base`. Are you saying that this PR could go father in abstracting the `base` content from the more form-specific content in `report_base`?

LindsayYoung:
Ah, thank you! I think I just got confused with the re-name.

LindsayYoung:
@el-mapache do you mind resolving the merge conflict? I want to make sure things get moved to the right place

el-mapache:
@LindsayYoung All fixed up!

PR #220, opened 2019-12-23
View complain details

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/186)

## What does this change?

This PR adds the `Complaint Details` card to the 'Complaint Details' page (meta!).

Screen Shot 2019-12-23 at 9 34 47 AM

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/views.py
+++ b/crt_portal/cts_forms/views.py
@@ -12,7 +12,7 @@ from django.conf import settings
 from formtools.wizard.views import SessionWizardView
 
 from .models import Report, ProtectedClass, HateCrimesandTrafficking
-from .model_variables import PROTECTED_CLASS_CODES
+from .model_variables import PROTECTED_CLASS_CODES, PRIMARY_COMPLAINT_CHOICES, HATE_CRIMES_TRAFFICKING_MODEL_CHOICES
 from .page_through import pagination
 from .filters import report_filter
 from .forms import Filters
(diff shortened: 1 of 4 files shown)


crt_portal/cts_forms/views.py
```diff
@@ -99,8 +99,21 @@ def IndexView(request):
...
     report = get_object_or_404(Report.objects, id=id)
+    primary_complaint = [choice[1] for choice in PRIMARY_COMPLAINT_CHOICES if choice[0] == report.primary_complaint]
+    crimes = {
+        'physical_harm': False,
+        'trafficking': False
+    }
+
+    for crime in report.hatecrimes_trafficking.all():
+        for choice in HATE_CRIMES_TRAFFICKING_MODEL_CHOICES:
+            if crime.hatecrimes_trafficking_option == choice[1]:
```
el-mapache: This feels brittle to me because it is based on the order of the content of each tuple being the same. I'm wondering if in a future / follow up PR, it makes more sense to store a reference to the type of crime in the report, and use that to match against the hate crime description.

For example, in the `protected_class` field, we store `disability`, not the string that choice maps to.

LindsayYoung: We can simplify this a bit. We should be able to see if an object is in a list or set without a loop. 

We might be able to use the option choice name instead of the text. We do that in the section assignment method I can't remember if it only works as a method or if it works outside the class too. 
```
If 'physical_harm' in report.hatecrimes_trafficking.all():
    crimes['physical_harm'] = True
``

el-mapache: Ah cool, if we can do this lookup I'd prefer that to looping and matching on the string!

PR #223, opened 2019-12-26
Route reports to IER

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/73)

## What does this change?

Adds assigned section routing to IER when the primary complaint involves the workplace, and at least one of the three protected classes outlined in the issue is indicated by the user.

### Author

### Reviewer

### DIFF
--- a/crt_portal/cts_forms/models.py
+++ b/crt_portal/cts_forms/models.py
@@ -108,10 +108,25 @@ class Report(models.Model):
     def __str__(self):
         return f'{self.create_date} {self.violation_summary}'
 
+    def __has_immigration_protected_classes(self, pcs):
+        immigration_classes = [
+            'Immigration/citizenship status (choosing this will not share your status)',
+            'National origin (including ancestry and ethnicity)',
+            'Language'
+        ]
+        is_not_included = set(pcs).isdisjoint(set(immigration_classes))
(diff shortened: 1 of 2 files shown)


crt_portal/cts_forms/models.py
```diff
@@ -108,10 +108,24 @@ class Report(models.Model):
     def __str__(self):
         return f'{self.create_date} {self.violation_summary}'
 
+    def __has_immigration_protected_classes(self, pcs):
+        immigration_classes = [
+            'Immigration/citizenship status (choosing this will not share your status)',
+            'National origin (including ancestry and ethnicity)',
+            'Language'
+        ]
+        for pc in pcs:
```
LindsayYoung: Instead of a loop, we can check if there is overlap directly. I feel like it is generally better to avoid loops, though, we know this will be a small loop, so tell me if you think this would be significantly harder to read, in that case it would not be worth it. 
```
is_distinct = set(pcs).isdisjoint(set(immigration_classes))
if is_distinct == True:
    return False
return True
```

el-mapache: My preference is definitely to use abstractions the language provides! I will make the change.

PR #232, opened 2020-01-03
Filter by complainant contact name

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/210)

## What does this change?

This PR adds the ability to filter complaints received by the complainant's first name, last name, or both. Note that the name filters, like each of the other individual filters, are also ANDs. That is, a search with `contact_last_name='smith'` and `contact_first_name='debbie'` will search for users with the name 'debbie smith' not `first_name=debbie OR last_name=smith`.

The name filters behave the same as the existing filters, that is, they produce tags with their values which can be clicked to remove that filter and requery the data.

Screen Shot 2020-01-03 at 2 12 48 PM

Screen Shot 2020-01-03 at 2 13 19 PM

### Author

### Reviewer

## Notes for reviewer:

We decided to make this two fields as combining both fields into a single input on the front-end was not trivial. Further user testing is needed to see if two fields are better than one!

### DIFF
--- a/crt_portal/cts_forms/filters.py
+++ b/crt_portal/cts_forms/filters.py
@@ -11,8 +11,8 @@ filter_options = {
     'status': '__in',
     'location_state': '__in',
     'primary_complaint': '__in',
-    'contact_first_name': '__search',
-    'contact_last_name': '__search',
+    'contact_first_name': '__contains',
+    'contact_last_name': '__contains',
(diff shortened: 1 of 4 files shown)


LindsayYoung:
Looks good! I updated the ACs to reflect first and last name being separate since I see from @Jacklynn's note that that change was approved.

PR #233, opened 2020-01-04
Small refactor of complaint viewer filter JS code

el-mapache:
## What does this change?

Removes duplication and cleans up some of the code around building a new query string of filters based on existing filter query parameters in the URL and user interaction with filtering form controls on the `/form/view` intake page.

### DIFF
--- a/crt_portal/static/js/complaint_view_filters.js
+++ b/crt_portal/static/js/complaint_view_filters.js
@@ -1,23 +1,4 @@
 (function(root, dom) {
-  var SEARCH_PARAMS_WHITELIST = ['sort', 'page', 'per_page'];
-  var filterData = {
-    assigned_section: [],
-    primary_complaint: '',
-    status: '',
-    location_state: '',
-    primary_complaint: '',
-    contact_first_name: '',
-    contact_last_name: '',
-    contact_email: '',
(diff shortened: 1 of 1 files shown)


LindsayYoung:
If I am understanding correctly, this will eliminate the need for creating the filter lists in the python view since it is parsing the query from to url. I would like to remove any unused code from there.

I am also feeling like we have hit the point where we need some js tests for this. @el-mapache do you have any feelings on this?

I haven't looked at this code before so I am still reading it over and testing.

el-mapache:
@LindsayYoung I don't think we need to remove any python code, this code just formats the form data into the query params that django needs to read. Testing seems like a good idea, but at the moment I'm not sure how much work it will be to introduce them. Additionally, we really only have 1-2 files that would benefit from it at this point (this one and the word count functionality). 

Also, I'm happy to go over any of the code with you!

LindsayYoung:
@el-mapache I am having an issue locally where it is taking off the last filter instead of the filter I clicked. 

Add all the filters. Try to remove ADM, then you VOT is removed instead. Does this happen to you?

el-mapache:
@LindsayYoung Yep, looks like a regression from the merge. It's a quick fix

PR #238, opened 2020-01-08
Show conditional location form when primary complaint is related to the police

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/198)

## What does this change?
This PR adds a conditional set of questions to the `Location` form step when the user indicates that their primary complaint is related to mistreatment by law enforcement.

* Adds two new model fields to capture:
  * Whether or not the alleged violation occurred inside a correctional facility
  * If it did, what type of correctional facility was it?

* Adds additional logic to conditionally display the extra form questions

Screen Shot 2020-01-09 at 3 56 21 PM

Screen Shot 2020-01-16 at 3 11 14 PM

Screen Shot 2020-01-16 at 3 05 01 PM

### Author

### Reviewer

## Notes for reviewer:

**Additional Notes**:

* I added a new template where I eschewed questions groups etc and just referenced the raw fields. This leads to duplication but is easier than coaxing existing objects into new forms. I think we may be at the point where we might want to turn the `QuestionGroup` class into a new abstraction (or set of abstractions).

* I basically had to duplicate the `other-show-hide` js in this new template, since that code is coupled to the `protect class` question set. This seems to point in the direction of some new widgets (or other classes) that can easily be decorated with client-side functionality.

* Is it time to start having more than 1 view? I don't think we need to have a single view that serves every form page, especially as we start adding more custom behavior to them. Certainly, at least in my opinion, the `Location` pages at deserve consideration for their own view. I think we could have 2 levels of inheritance without the codebase getting too crazy i.e. 
```SessionWizardView -> CRTReportWizard -> LocationView```

* The pa11y tests are failing again, these seem brittle. I almost think we might want to look at using a different command line program to run them. They are super hard to debug and all errors don't get displayed when tests fail e.g. [see](https://github.com/pa11y/pa11y-ci/issues/73)

### CODE REVIEW

File: crt_portal/cts_forms/migrations/0030_auto_20200108_1928.py [hunk 1]
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.2.8 on 2020-01-08 19:28
```
LindsayYoung: Small thing- do we want to rename this file something descriptive?

crt_portal/cts_forms/templates/forms/report_police_location.html
```diff
@@ -0,0 +1,38 @@
...
+    <div class="other-class-option padding-left-4 margin-bottom-2 padding-top-3">
+      <div class="blue-left-highlight">
+        <label for="{{ field.id_for_label }}">
+          <b>{{ field.help_text }}</b>  
+        </label>
+        {{ field }}
+      </div>
+    </div>
+  {% endwith %}
+
```
LindsayYoung: We need the error for the field here. The required error is not showing up under the question.

crt_portal/cts_forms/migrations/0030_auto_20200108_1928.py
el-mapache: Definitely!!

el-mapache: So, I actually can't get this file renamed without producing a bunch of database errors, specifically: 

```
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration cts_forms.0031_merge_20200116_1618 is applied before its dependency cts_forms.0030_add_correction_facility_fields on database 'default'.
```

I couldn't exactly track down how to fix this; if you have ideas, I would love to hear em 😄

LindsayYoung: We have to change the dependencies reference in the next file to match with the new name, not 100% sure it is worth it right now

LindsayYoung:
I noticed that on page 4 and that the other reason box is now expanded and can't be unexpanded.

I am getting a data error on save- that might be me, I will keep looking.

el-mapache:
@LindsayYoung You might need to rebuild the css; the other checkbox is togglable on my local version of this branch

LindsayYoung:
Woo Hoo!!  This is good to go

PR #266, opened 2020-01-28
Conditionally route complaints to SPL

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/92)

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/tests.py
+++ b/crt_portal/cts_forms/tests.py
@@ -376,6 +376,39 @@ class SectionAssignmentTests(TestCase):
         self.assertFalse(test_report.assign_section() == 'EOS')
         self.assertTrue(test_report.assign_section() == 'ADM')
 
+    def test_SPL_routing(self):
+        disability = ProtectedClass.objects.get_or_create(protected_class='Disability (including temporary or recovery)')
+
+        # Test if law enforcement and inside a prison
+        data = copy.deepcopy(SAMPLE_REPORT)
+        data['primary_complaint'] = 'police'
+        data['inside_correctional_facility'] = 'inside'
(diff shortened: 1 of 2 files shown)


crt_portal/cts_forms/tests.py
```diff
@@ -376,6 +376,39 @@ def test_education_routing(self):
...
+        # Test if law enforcement and inside a prison
+        data = copy.deepcopy(SAMPLE_REPORT)
+        data['primary_complaint'] = 'police'
+        data['inside_correctional_facility'] = 'inside'
+        test_report = Report.objects.create(**data)
+        self.assertTrue(test_report.assign_section() == 'SPL')
+
+        test_report.protected_class.add(disability[0])
+        test_report.save()
+        self.assertTrue(test_report.assign_section() == 'ADM')
```
LindsayYoung: Do we want check not 'SPL' rather that checking for ADM, since I think that routing will change?

el-mapache: I'm not sure? The routing hasn't changed in this ticket, so for now it will still go to `ADM` when the conditions aren't met, right? I think perhaps I'm misunderstanding something

LindsayYoung: The disability ones are going to Disability once we implement that ticket. So, we can fix the texts then, or make them more forgiving now. It's not a non starter if you want to leave them as is.

el-mapache: Ah gotcha, wasn't sure how far out the DRS routing was! I'll change it to the absence of SPL so we dont have to update it again!

PR #267, opened 2020-01-29
[Don't merge yet] Release/02-04-20

LindsayYoung:


LindsayYoung:
Closing in favor of: https://github.com/usdoj-crt/crt-portal/pull/298

Good news is that now we have SSO going, we don't need prod for testing that and we can stay on  our regular release schedule!

PR #269, opened 2020-01-29
Adds status select field to complaint detail view

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/248)

## What does this change?

* Adds `select` element to complaint details view with available `status` values
* Extends ComplaintSelect widget to allow for custom labeling
* Styles will not match mocks yet

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/widgets.py
+++ b/crt_portal/cts_forms/widgets.py
@@ -23,6 +23,25 @@ class ComplaintSelect(ChoiceWidget):
     template_name = '../templates/forms/widgets/complaint_select.html'
     option_template_name = '../templates/forms/widgets/multi_select_option.html'
 
+    def __init__(self, *args, **kwargs):
+        label = kwargs.pop('label', None)
+
+        ChoiceWidget.__init__(self, *args, **kwargs)
+
+        self.label = label
+
+    def label_for_widget(self):
+        return self.label
+
+    def render(self, name, value, attrs=None, renderer=None):
+        extra_context = {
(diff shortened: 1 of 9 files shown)


crt_portal/cts_forms/widgets.py
```diff
@@ -23,6 +23,25 @@ class ComplaintSelect(ChoiceWidget):
...
+        return self.label
+
+    def render(self, name, value, attrs=None, renderer=None):
+        extra_context = {
+            'label': self.label_for_widget()
+        }
+        context = self.get_context(name, value, attrs)
+        context.update(extra_context)
+
+        return self._render(self.template_name, context, renderer)
```
el-mapache: I directly call render here, but it is probably better / safer to call super().render(...args)?

el-mapache: Calling super here seems to open a large can of worms, so I wonder if maybe this is ok for now 😬

LindsayYoung:
Small thing- if you change the status or section, it doesn't preserve the link in the close button.

el-mapache:
@LindsayYoung I've updated the code, but I wonder if there is a cleaner (or more Django-specific) way to achieve this logic?

LindsayYoung:
> @LindsayYoung I've updated the code, but I wonder if there is a cleaner (or more Django-specific) way to achieve this logic?

I am thinking we can add a hidden field that passes the next variable.

el-mapache:
works for me!

PR #319, opened 2020-02-26
Add Report field tracking modification times and display on details page

Jkrzy:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/252)

## What does this change?
Herein we add `modified_date` to the `Reports` model which will be updated with the current timestamp every time a `Report` instance is updated via the application. We display this timestamp on the Reports details page.

### Behind the scenes

This is a Django feature which automatically updates the value when a model's `save` method is called. Following common Django patterns, this should occur every time an instance is modified via the app. This PR includes a bit of refactoring of `ShowView.post` to leverage built-in Django functionality to more concisely validate/process the incoming data and effect the associated changes to the Report.

Scenarios where this date field will not be automatically updated:

- Changes applied directly to the database
- A `Report` instance is modified in a manner that does not call `.save().` e.g. QuerySet.update()

These situations should be rare and if needed they must explicitly set the date modified as part of implementation. 

Also includes a minor bug fix, the timestamps on the details page were previously displaying `hours:seconds` instead of `hours:minutes`.  

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/migrations/0049_fix_other_duplicates.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('cts_forms', '0046_update_primary_complaint'),
+        ('cts_forms', '0050_auto_20200226_1532'),
```
LindsayYoung: I am getting an error running this migration.
```
raise InconsistentMigrationHistory(
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration cts_forms.0049_fix_other_duplicates is applied before its dependency cts_forms.0050_auto_20200226_1532 on database 'default'.
```
 I don't think my DB and the dev DB that have already applied these [migrations up to migration 50](https://github.com/usdoj-crt/crt-portal/tree/develop/crt_portal/cts_forms/migrations), so I don't think we can change the order on the migrations that were already run. I think if you keep the order as it is in the develop branch, have your new migration fork from where you were, a new merge migration could get them synced up.  

Happy to connect if you want to pair on this tomorrow.

LindsayYoung:
Test folder is a nice cleanup idea! let's move the other test and test data to the test folder and make sure we update and path examples to run the tests if we need to.  

Breaking up the test files some more sounds good as well, but we might want to break them up by form tests and crt view tests (or some other heuristic). The biggest lift on that will be disentangling the imports, but we could use the flake8 checker to help us out.

One small thing is that we have been renaming the migrations something descriptive.

I am impressed with how quickly you are jumping in and finding your way around!

crt_portal/cts_forms/migrations/0049_fix_other_duplicates.py
Jkrzy: Ah thanks, will fix this morning!

Jkrzy: Ooo now I'm remembering, the problem I was trying to solve here is a conflict between `0049` and adding `modified_date`.

The data migration in `0049` triggers the `auto_now` and attempts to write to `Report.modified_date` before it exists.

crt_portal/cts_forms/migrations/0051_auto_20200227_1408.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.9 on 2020-02-27 14:08
```
LindsayYoung: Pretty sure this file was just renamed so it should just be deleted.

Jkrzy: Oops, good catch. Thanks! Deleted

PR #320, opened 2020-02-26
Updates styles for view all page

el-mapache:
[Link to ZenHub issue.](https://github.com/18F/crt-portal/issues/231)

## What does this change?

Large overhaul of the view all table view on the CRT intake side. This PR brings the intake view more inline with our current design vision, and introduces a number of changes to the structure of the view all table's html + js.

* Removes unused styles (filters.scss)
* Adds filter html template that individual filter components can
      extend. Filter.html uses new crt-dropdown component
* Adds JS to handle dropdown behavior. Similar to uswds accordion,
      but not tied to the uswds accordion classes
* Minor tweaks to the filter JS to allow us to clear filters
      Now, instead of directly mutating the initial filter state,
      we copy the initial state and mutate the copy. This allows us
      to restore the filter model to its defaults before sending the new query
      to the backend
* Fixes extra margin above `service member` question in public portal

Screen Shot 2020-02-27 at 12 16 05 PM

**Header**
Screen Shot 2020-02-26 at 10 08 56 AM

**Filters**
Screen Shot 2020-02-27 at 11 06 10 AM
Screen Shot 2020-02-27 at 11 06 22 AM
Screen Shot 2020-02-27 at 12 15 16 PM

**Table**
Screen Shot 2020-02-27 at 11 06 31 AM

### Author

### Reviewer

## Notes for reviewer:


LindsayYoung:
@el-mapache This looks so fantastic!

In poking around I found one little visual thing:

- If I look at it in 90% alignment is correct. At actual size in chrome I am seeing an alignment issue where the table spills over on the right.

Screen Shot 2020-02-28 at 8 59 22 AM

### pa11y
Ran the pa11y test locally. (it needs to be updated to have the trailing slash, I can add that in another PR) Here are the errors it returned:

Errors in http://0.0.0.0:8000/form/view:

 • This select element does not have a name available to an accessibility API. Valid names are: label
   element, title undefined, aria-label undefined, aria-labelledby undefined.

   (#id_assigned_section)

   <select class="complaint-multi-select text-uppercase" multiple="" name="assigned_section"
   classes="text-uppercase" id="id_assigned_section"> <option val...</select>

 • This form field should be labelled in some way. Use the label element (either with a "for" attribute or
   wrapped around the form field), or "title", "aria-label" or "aria-labelledby" attributes as appropriate.

   (#id_assigned_section)

   <select class="complaint-multi-select text-uppercase" multiple="" name="assigned_section"
   classes="text-uppercase" id="id_assigned_section"> <option val...</select>

 • Duplicate id attribute value "page-forward" found on the web page.

   (#page-forward)

   <span id="page-forward" class="usa-sr-only">Next page.</span>


### Notes on enhancements we might want to do but aren't needed for this PR:

- have a blank default for the state dropdown. Right now it looks like you are searching for Alabama even if you aren't. To search for Alabama, you need to click a another state and then back on Alabama.
Screen Shot 2020-02-28 at 9 04 38 AM
- we might want multi-select for state in the future

el-mapache:
@LindsayYoung That default is being added in [this PR](https://github.com/usdoj-crt/crt-portal/pull/325) that @Jkrzy put in!

el-mapache:
@LindsayYoung Can you tell me a little more about this issue? I am developing the view with Chrome at 100% zoom, and I don't actually see any overflow. We will definitely get overflow if the page is made significantly smaller, since tables aren't natively responsive 😢

PR #327, opened 2020-02-28
update remaining images for deploy

LindsayYoung:
All images need to be on 3.8.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,7 +115,7 @@ jobs:
     working_directory: ~/code
     docker:
       # standard docker python image
-      - image: python:3.8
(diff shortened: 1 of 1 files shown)


LindsayYoung:
> Do we need to update the `paths` values here as well? e.g. [4da79f2](https://github.com/usdoj-crt/crt-portal/commit/4da79f265ae9cca215a93a82d4251c2d0bf1e050)

I don't think so, because this doesn't change what version of python we are using. In the other PR we were upgrading the python version.

LindsayYoung:
Ok, since you approved, I can merge and we can see if I am wrong about the paths

PR #334, opened 2020-03-04
Pro form

LindsayYoung:
[Proform](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/332)

## What does this change?
- Adds a pro-form with all the current questions.
- removes unused form fields

crt-new

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/models.py
```diff
@@ -81,59 +81,75 @@ class Report(models.Model):
...
-    public_or_private_healthcare = models.CharField(max_length=100, null=True, choices=PUBLIC_OR_PRIVATE_HEALTHCARE_CHOICES, default=None)
+        try:
+            day = self.last_incident_day or 1
+            date = datetime(self.last_incident_year, self.last_incident_month, day)
+        except ValueError:
+            date = None
+        return date
+
+    @cached_property
+    def crt_reciept_date(self):
```
Jkrzy: I don't think these functions, `last_incident_date` and `crt_reciept_date` were actually modified here but noting that they don't appear to be call anywhere in the current code base. Potential opportunity to remove them here.

crt_portal/cts_forms/signals.py
```diff
@@ -42,10 +42,15 @@ def format_data_message(action, current_request, instance):
...
-    logger.info(message)
+def save_user(sender, instance, created, **kwargs):
+    if created:
+        current_request = CrequestMiddleware.get_request()
+        message = format_user_message('created', current_request, instance)
+        logger.info(message)
+    else:
+        current_request = CrequestMiddleware.get_request()
+        message = format_user_message('saved', current_request, instance)
+        logger.info(message)
```
Jkrzy: Great addition to differentiate this log output! Opportunity here to use a ternary and cut a bit of the duplication with something like:

```python
def save_user(sender, instance, created, **kwargs):
    action = 'created' if created else 'saved'
    current_request = CrequestMiddleware.get_request()
    message = format_user_message(action, current_request, instance)
    logger.info(message)
```

crt_portal/cts_forms/views.py
```diff
@@ -178,6 +178,74 @@ def post(self, request, id):
         return render(self.request, 'forms/complaint_view/show/index.html', output)
 
 
+class ProFormView(SessionWizardView):
+    @method_decorator(login_required)
+    def dispatch(self, *args, **kwargs):
```
Jkrzy: Opportunity here to leverage Django's [`LoginRequiredMixin`](https://docs.djangoproject.com/en/3.0/topics/auth/default/#the-loginrequired-mixin) instead of overriding `dispatch`

crt_portal/cts_forms/views.py
```diff
@@ -178,6 +178,74 @@ def post(self, request, id):
         return render(self.request, 'forms/complaint_view/show/index.html', output)
 
 
+class ProFormView(SessionWizardView):
```
Jkrzy: Since we're not using a multi-stage form, could this be simpler as a regular old `FormView`?

crt_portal/cts_forms/views.py
```diff
@@ -178,6 +178,74 @@ def post(self, request, id):
...
+            o = HateCrimesandTrafficking.objects.get(hatecrimes_trafficking_option=option)
+            r.hatecrimes_trafficking.add(o)
+
+        r.assigned_section = r.assign_section()
+        r.save()
+        # adding this back for the save page results
+        form_data_dict['protected_class'] = m2m_protected_class.values()
+        form_data_dict['hatecrimes_trafficking'] = m2m_hatecrime.values()
+
+        return HttpResponseRedirect(f'/form/view/{r.pk}')
```
Jkrzy: This looks to be largely a duplicate of `CRTReportWizard.done`, good opportunity to pull this custom save logic into it's own function or maybe `Report.save` and call where needed.

Also a great spot and opportunity to add a [get_absolute_url](https://docs.djangoproject.com/en/3.0/ref/models/instances/#get-absolute-url) function to `Report` and use that here. We could definitely use that elsewhere.

crt_portal/cts_forms/forms.py
```diff
@@ -563,8 +577,50 @@ def __init__(self, *args, **kwargs):
...
+        # a bit of a catch-all for all the ways people could make bad dates
+        self.add_error('last_incident_year', ValidationError(
+            _(f'Invalid date format {month}/{day}/{year}.'),
+            params={'value': f'{month}/{day}/{year}'},
+        ))
+    except KeyError:
+        # these will be caught by the built in error validation
+        return cleaned_data
+
+    return cleaned_data
```
Jkrzy: It _may_ be easier to handle and test the `day` and `year` checks as field level clean functions e.g. `clean_last_incident_year` and leave the ones that _must_ look across fields to this function which is called by `clean` in a few places.

crt_portal/cts_forms/forms.py
```diff
@@ -668,6 +687,219 @@ class Meta:
...
+    Contact,
+    HateCrimesTrafficking,
+    ElectionLocation,
+    WorkplaceLocation,
+    CommercialPublicLocation,
+    PoliceLocation,
+    EducationLocation,
+    ProtectedClassForm,
+    When,
+):
```
Jkrzy: Great re-use of existinng forms!

crt_portal/cts_forms/forms.py
```diff
@@ -668,6 +687,219 @@ class Meta:
...
+
+        self.fields['crt_reciept_day'].label = DATE_QUESTIONS['last_incident_day']
+        self.fields['crt_reciept_month'].label = DATE_QUESTIONS['last_incident_month']
+        self.fields['crt_reciept_year'].label = DATE_QUESTIONS['last_incident_year']
+
+        self.fields['violation_summary'].widget.attrs['class'] = 'usa-textarea word-count-500'
+        self.label_suffix = ''
+        self.fields['violation_summary'].label = SUMMARY_QUESTION
+        self.fields['violation_summary'].widget.attrs['aria-describedby'] = 'details-help-text'
+        self.fields['violation_summary'].help_text = 'What did the person believe happened?'
```
Jkrzy: This works but I _think_ there's a good spot here to tackle the duplication of widget/field definitions across `forms.py`. A bigger effort that would span across most of the other Forms defined here -- let's tackle it in a separate issue.

crt_portal/cts_forms/forms.py
```diff
@@ -668,6 +687,219 @@ class Meta:
...
+
+    def clean(self):
+        """Validating more than one field at a time can't be done in the model validation"""
+        cleaned_data = super(ProForm, self).clean()
+        if cleaned_data['last_incident_year'] is not None and cleaned_data['last_incident_month'] is not None:
```
Jkrzy: I believe these are required fields, can we evaluate this with?
```python
if cleaned_data['last_incident_year'] and cleaned_data['last_incident_month']:
```

crt_portal/cts_forms/templates/forms/report_date.html
```diff
@@ -1,56 +1,14 @@
...
-        <legend class="em-text ">
-          {{ wizard.form.date_question }}
-        </legend>
+<div class="crt-portal-card">
+  <div class="crt-portal-card__content crt-portal-card__content--lg">
+    {{ block.super }}
+
+    <h2>{{ current_step_title }}</h2>
+
+    {% include "forms/question_cards/date.html" with question=wizard.form.date_question month=wizard.form.last_incident_month day=wizard.form.last_incident_day year=wizard.form.last_incident_year help_text=wizard.form.help_text %}
```
Jkrzy: `wizard.form` and `form` are equivalent in the context here, good spot to trim a bit and reference `form` directly

e.g.
```python
    {% include "forms/question_cards/date.html" with question=form.date_question month=form.last_incident_month day=form.last_incident_day year=form.last_incident_year help_text=form.help_text %}
```

crt_portal/cts_forms/templates/forms/report_base.html
```diff
@@ -14,13 +14,15 @@
     <div class="grid-row grid-gap">
       <div class="tablet:grid-col-8 tablet:grid-offset-2">
         {% block page_info_card %}{% endblock %}
+        {% block form_id %}
         <form id="report-form"
               class="usa-form"
               action="/report/"
```
Jkrzy: Not changed in this PR but noting here that this is a good spot to use [{% url %}](https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#url) as opposed to hard coding the path.

crt_portal/cts_forms/templates/forms/question_cards/single_question.html
```diff
@@ -0,0 +1,19 @@
+<div class="form-group">
+  {% if field.label %}
+    <h3><label
+      for="{{ field.id_for_label }}"
+      class="{{label_class}} margin-bottom-0"
+    >
+      {{ field.label }} {% if field.field.required != True %}<em>(Optional)</em>{% endif %}
```
Jkrzy: Small thought, we could handle this conditional with `if not field.field.required`

crt_portal/cts_forms/templates/forms/pro_template.html
```diff
@@ -0,0 +1,99 @@
...
+<a name="Intake"></a>
+<div class="crt-portal-card">
+  <div class="crt-portal-card__content crt-portal-card__content--lg">
+        {% if page_errors %}
+          <section id="page-errors" class="display-none page-errors margin-bottom-2" role="alert" aria-live="assertive">
+            {{ num_page_errors }} error{{ num_page_errors|pluralize }}: {{ page_errors_desc }}
+          </section>
+        {% endif %}
+
+        <a name="Intake"></a>
```
Jkrzy: We could replace these two duplicated and empty anchor tags w/ an `id="Intake"` attribute on the element most relevant for each section. The `h2` could be appropriate here.

crt_portal/cts_forms/models.py
LindsayYoung: `last_incident_date` was required and now isn't required for the pro form. We will need that when we add date range sort on the incident date. I try to front-load model changes as much as possible, because I feel like doing a lot of little model changes across a few people can get harry fast. 

`crt_reciept_date` is a new optional field for the pro form, but there isn't currently a place for its display on the back end. The use case for this info is to figure out how much time it take DOJ to get the mail to CRT. A date filed can be queried for those kinds of questions in a way that is easier than querying the month day and year fields. So the data will be available and packaged for data calls even if there isn't a nice way to get the data now. I have the related day, month and year fields just to keep parody with how we are inputing and processing the date fields for the front end.

crt_portal/cts_forms/views.py
LindsayYoung: Ok, I removed the wizards, there is a bit more dependence on the wizard extras, like steps and management forms in the base template. My instinct is to leave it as is, otherwise, we are pushing the complexity to the templates that would have to account for forms being provided in two different ways.  We can talk about it more later, if you like.

crt_portal/cts_forms/forms.py
LindsayYoung: Sounds good, I was trying to do some of that but ran into some issues because I couldn't pass attributes the same way for our custom widgets and the standard widgets. I would love to have some time to clean that up a bit!

crt_portal/cts_forms/forms.py
LindsayYoung: IIRC, the day and the year validation from the model won't get called because the date will fail in this clean method first and then it will give the generic bad date message. So we add the day, month, year messages here so that the most specific message will appear.

LindsayYoung:
Replaced anchors with ids. It should be good to go!

PR #344, opened 2020-03-12
[WIP/Don't Merge] Report comments

Jkrzy:
Went a little bit down the path on https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/6

Pushing that rough WIP here. 

**Quick thoughts, from my current understanding, on comments:**
We're using [django-activity-stream](https://github.com/justquick/django-activity-stream) for tracking report updates.

Need to decide if we're going to store comments as their own model or [data on the activity stream](https://django-activity-stream.readthedocs.io/en/latest/data.html). At the moment I'd lean towards the latter for ease of integration into the existing activity stream.


crt_portal/cts_forms/models.py
```diff
@@ -222,3 +217,17 @@ def assign_section(self):
...
         return 'ADM'
+
+
+class Comment(models.Model):
+    """
+    Store annotations/comments from users for
+    Reports
+    """
+    report = models.ForeignKey('Report', on_delete=models.CASCADE, editable=False, related_name="comments")
+    text = models.TextField(editable=False)
```
Jkrzy: Should be editable :oops:

Jkrzy:
replaced by #345

PR #355, opened 2020-03-18
Update errors

LindsayYoung:

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/forms.py
```diff
@@ -587,41 +592,36 @@ def __init__(self, *args, **kwargs):
...
-                _('Please enter a year after 1900.'),
-                params={'value': test_date.strftime('%x')},
+                DATE_ERRORS['no_past'],
+                params={'value': datetime(year, month, day).strftime('%x')},
             ))
 
     except ValueError:
-        # a bit of a catch-all for all the ways people could make bad dates
+        # a bit of a catch-all for all the ways people could make invalid dates
         self.add_error('last_incident_year', ValidationError(
```
Jkrzy: The `ValueError` provided by datetime is pretty descriptive, may be worth including here appended to `DATE_ERRORS['not_valid']`

i.e. 2/29/2019 passed to `datetime` will throw an exception of `Error: day is out of range for month`

LindsayYoung: Agreed that there is some room for improvement here.  I think we can circle back on it. 

These were the messages that were asked for and I know they put a lot of thought in to making the content for to a wide audience, especially people from vulnerable populations that may not have as much exposure to math terms.  In this example, I am not sure what percentage of the population we are trying to reach would understand what a range is.  

My general preference would be to not write custom messages, especially for dates because there are so many ways to screw up dates. (Having the day optional also make the automatic checking more complicated.) 

But, the good thing about the custom messages, is that they will get translated when we add internationalization, so it might be worth the custom messages in the end anyways.

Jkrzy: Really good point about the internationalization!!

PR #358, opened 2020-03-20
Load testing script

LindsayYoung:
[Load Testing Production#64](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/64)

## What does this change?
For the test we will simulate user behavior:

- submissions via /report
- submissions via /report/new
- data queries on /form/view
- views of the detail page at /form/view/id
- posting of comments

Screen Shot 2020-03-20 at 4 53 01 PM

### Author

### Reviewer

## Notes for reviewer:


load_testing/locust.py
```diff
@@ -0,0 +1,142 @@
...
+        pro_form: 3,
+        get_report: 10,
+
+    }
+
+    def on_start(self):
+        login(self)
+
+    def on_stop(self):
+        logout(self)
```
Jkrzy: Is there a low-effort way here to differentiate between anonymous and authenticated users? This will have every `locust` attempting to login and logout

LindsayYoung: There is a way to do that, but I don't think it would be easy because I think the tasks are assigned at random in the proportion that you designate. Of course, I am open to suggestions. 

The hatch rate should keep them from logging in all at once. They will all log out at once.

LindsayYoung:
I am thinking we can merge this and revisit once we do another round of testing once DNS is set up. Sound good @Jkrzy ?

PR #366, opened 2020-03-27
Release/03 27 2020

LindsayYoung:
Calling out that there are 3 small changes 
- no longer binds app to log drain we are not using
- removes specified worker limit
- increases production instances to 4


LindsayYoung:
This was deployed last Friday. Merging to master to make sure code is up to date and confirming we are now properly resourced in cloud.gov

PR #381, opened 2020-04-07
Load zips from S3

LindsayYoung:
Fixes the build.
We were exceeding our memory limit and this was a large file that makes sense to save as data on S3 rather than in the codebase. 

## What does this change?
removes the zip code file, points to a cloud hosted zip code file. 

(You can test this by adding a print statement in migration 61 and then running a test. you can have the data in your terminal.)

This is running for the tests, but I have code to skip those, but it is on a different branch. Though, this makes testing the import easy.

I confirmed this is in our disk limit for deploy. 

### Author

### Reviewer

## Notes for reviewer:


LindsayYoung:
@Jkrzy will you get a chance to look at this one today? I could also merge it for now to get the deployment pipeline going and then you could do a more through review later, if that would work better fo you.

LindsayYoung:
@Jkrzy I just realized you reviewed translation- which accidentally had this in there.

PR #382, opened 2020-04-08
Translation

LindsayYoung:
[provide content for translation](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/436)

## What does this change?

- We were pretty good about having the translations in .py files but we missed a lot of text in the templates. This adds that text. 
- Clearly marked the backend content that did not need to be translated.
- Creates the .po .mo files 
- I made a .txt version for translators that is on the zenhub ticket

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/migrations/0061_load_zips.py
```diff
@@ -1,13 +1,24 @@
...
 
 logger = logging.getLogger(__name__)
 
-circle = os.environ.get('CIRCLE', False)
+
+def fetch_file():
+    # no credentials needed to pull from the public bucket
+    s3 = boto3.resource('s3', config=Config(signature_version=UNSIGNED), region_name='us-gov-west-1')
+    # this is a public link in the dev s3 bucket
+    bucket = s3.Bucket(u'cg-b2626a97-5c90-4d35-a39b-24f3b035d5c9')
```
Jkrzy, line 19: Not a show stopper but there's an opportunity to define this in settings from an environment variable

crt_portal/cts_forms/migrations/0061_load_zips.py
```diff
@@ -1,13 +1,24 @@
...
 import os
-from csv import DictReader
+
 
 from django.conf import settings
 from django.db import migrations, models
 
 logger = logging.getLogger(__name__)
 
-circle = os.environ.get('CIRCLE', False)
```
Jkrzy, line 10: Was dropping the check for test environments intentional here? We still have the comment below about not needing to run this under test.

crt_portal/cts_forms/templates/forms/base.html
```diff
@@ -1,4 +1,5 @@
 {% load static %}
+{% load i18n %}
 {% load with_input_error %}
 <!DOCTYPE html>
 <html lang="en">
```
Jkrzy, line 5: We're not implementing the translation here but flagging since we'll need to update this in a new issue.

crt_portal/cts_forms/templates/forms/confirmation.html
```diff
@@ -94,60 +93,61 @@ <h3 class="intake-section-title">Our specialists determine the next step</h3>
...
 
     <div class="tablet:grid-col-8 tablet:grid-offset-2">
-      <h2>What you can do next</h2>
+      <h2>{% trans "What you can do next" %}</h2>
       <ol class='confirmation-list'>
         <li>
           <div class="title">
             <div class="title-icon-blue margin-right-2">
-              1
+              {% trans "1" %}
```
Jkrzy, line 122: Should these be translated? If yes, there's a few missing template tags int the section numbering above.

Jkrzy:
This looks _amazing_ @LindsayYoung, thanks for putting it together! Made a few small comments that are worth a quick look/response before merging, I don't expect there to be substantial changes, if any, from them.

In the interim, i'm taking a closer look and combing through the code now to double check we're not missing any other strings, will submit a PR for those if I find any or add-on here if we haven't merged yet.

Jkrzy:
Do we need to translate labels read by screen readers like here? I suspect yet, but am not sure...

https://github.com/usdoj-crt/crt-portal/blob/88da5ebffedf2de8bc74d6a196f8c150a9d56221/crt_portal/cts_forms/templates/forms/report_base.html#L66

crt_portal/cts_forms/templates/forms/confirmation.html
LindsayYoung: This PR is for Spanish, so the numbers are going to be the same. I can take them out for now if it is confusing.

I do believe that we will have to start adding the numbers for other languages- but haven't confirmed. I tried to find the hard coded numbers, since that was low hanging fruit but there are some calculated numbers that we will have to do some refactoring for in the future. 

For the epic: https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/201
The eventual languages will be:
- Spanish
- Korean
- Taglog
- Vietnamese
- Mandarin
- Simplified Chinese

crt_portal/cts_forms/migrations/0061_load_zips.py
LindsayYoung: Sorry, this was a stray commit from another issue. 

That variable isn't used. So it doesn't actually change anything. We only have dynamic variables from .env files and don't have per-environment args going for docker so there are environment variables during the build but not while the app is running.

Jkrzy:
Pushed ca8c533ec92ed92a1fa8f2fb95406263b7db2783 w/ two more strings for translation. 

Caught them with https://www.technomancy.org/python/pylint-i18n-lint-checker/ which needed some hacking to get up and running and isn't suitable at the moment for integration into the test suite but could be with some additional work. Helpful for spot checks at this point at least.

crt_portal/cts_forms/migrations/0061_load_zips.py
LindsayYoung: We would want the same variable in Prod, Staging, Dev and Local. There is one document that is the source of truth so we want everyone pointing there.  Prod, Staging and local don't have this variable defined and this is the only place that we will be using it. 

In another PR I remove the zip loading from tests and that speeds each run up by a minute. This will only be run for a brand new database, say developer on-boarding or if they want to make a new enviornment in a sandbox or in feature etc.

crt_portal/cts_forms/forms.py
```diff
@@ -906,7 +906,7 @@ def __init__(self, *args, **kwargs):
         self.label_suffix = ''
         self.fields['violation_summary'].label = SUMMARY_QUESTION
         self.fields['violation_summary'].widget.attrs['aria-describedby'] = 'details-help-text'
-        self.fields['violation_summary'].help_text = 'What did the person believe happened?'
+        self.fields['violation_summary'].help_text = _('What did the person believe happened?')
```
LindsayYoung, line 909: This is only on the CRT view and doesn't appear on the public web

crt_portal/cts_forms/templates/forms/base.html
LindsayYoung: Maybe I am misunderstanding, but I think this is the load step for {% trans %} tag we use later in the file.

LindsayYoung:
> Do we need to translate labels read by screen readers like those on the `Back` buttons? I suspect yet, but am not sure...
> 
> https://github.com/usdoj-crt/crt-portal/blob/88da5ebffedf2de8bc74d6a196f8c150a9d56221/crt_portal/cts_forms/templates/forms/report_base.html#L66

It seems screen reader do support other languages 🎉
https://accessibility.psu.edu/foreignlanguages/

So it seems like a good accommodation to make. I can do a pass at front end labels. Nice catch!

PR #397, opened 2020-04-14
Render report author on details page in activity stream

Jkrzy:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/427)

## What does this change?
Render an activity stream entry detailing the creation of each report.

We're appending this to the bottom of the activity stream
when rendering as opposed to entering into the stream itself.

This was an implementation decision made to accommodate
representing actions by actors not associated with a
User in the database. Specifically, anonymous public actors.

### Author

### Reviewer

## Notes for reviewer:


Jkrzy:
Whoops, fix for failing test incoming...

crt_portal/cts_forms/templates/forms/complaint_view/show/activity_stream.html
```diff
@@ -19,5 +19,16 @@
         </p>
       </li>
     {% endfor %}
+    <li class="activity-stream-item">
+      <div class="display-flex flex-justify">
+        <b>{% firstof data.author_user.email data.author  %}</b>
```
LindsayYoung, line 24: We don't show email in the activity feed, we are showing user name. So I think we either want to add the same logic there or make it username here.

Jkrzy: Agh good catch I didn't notice that inconsistency when implementing. 

The originating [issue](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/427) calls for email, and I feel like I'm missing some context around if `username` is a helpful identifier as-is or if email would be more helpful. We've got the code here for both approaches 🤷 

Who can make that call? 😄

Jkrzy: Per Lalitha, we'll continue with `username` across the board. Update incoming...

LindsayYoung: So the backstory on that- In theory we should get the email and name claims from DOJ SSO but it isn't working. We got sign on going and didn't really have more time to dig into it. Since username will always be populated that is what we went with showing. In the future, emails could be added manually or we could reach out to JMD to debug the claims issue.

LindsayYoung:
Looks great and now we have a good way to handle the public email info in the future!

PR #432, opened 2020-04-29
Implement translation

Jkrzy:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/202)

## What does this change?

Makes available public facing pieces of the portal in Spanish.

Included here:
* Hookup of Django settings/configuration to serve translated pages
* Inclusion of a temporary form in the footer of the page to change language choice for the current session.
* Addition of a handful of translation tags to handle untranslated pieces noticed on initial rollout
* Modified form logic for `ProtectedClasses` to display all existing objects as options instead of filtering by translated label
* Updated `messages.po` file including a few new strings which need translation

## TODO
There's a handful of areas that still need translation implementation I noticed while implementing this. To not drag the scope of this PR into infinity, we may want to tackle them in as separate issues. 

- [ ] Deliver updated `messages.po` file for updated/new translations. The [diff](2b08f9c1b2442a6c5174085714750e3bda59115c) has some unintentional changes to file paths, I need to regenerate this.
- [ ] Refine and finalize design and implementation of language selector form

Example: Error messages
Screen Shot 2020-04-29 at 4 16 13 PM

### Initial Form
Screen Shot 2020-04-29 at 4 15 38 PM

### Language selector form
Screen Shot 2020-04-29 at 4 15 46 PM

### Confirmation page
Screen Shot 2020-04-29 at 4 16 32 PM

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/templates/forms/report_data.html
```diff
@@ -53,15 +54,15 @@ <h3>{{ question.hatecrime_title }}</h3>
         <h4>{{ question.hatecrime }}</h4>
         {% if report.id %}
             {% for crime in report.hatecrimes_trafficking.all %}
-                <p>{{ crime }}</p>
+                <p>{% trans crime.hatecrimes_trafficking_option %}</p>
```
LindsayYoung, line 57: Shouldn't this be taken care of by lazy loading?

https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/model_variables.py#L75-L78

crt_portal/cts_forms/templates/forms/report_data.html
```diff
@@ -156,15 +157,15 @@ <h2>{{ ordered_step_names.3 }}</h2>
         <h3>{{ question.characteristics }}</h3>
         {% if report.id %}
             {% for characteristic in report.protected_class.all %}
-                <p>{{ characteristic }}</p>
+                <p>{% trans characteristic.protected_class %}</p>
```
LindsayYoung, line 160: I think this should be taken care of here:
https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/model_variables.py#L84-L103

crt_portal/cts_forms/templates/forms/widgets/usa_checkbox_option.html
```diff
@@ -6,5 +7,5 @@
     class="usa-checkbox__input"
     {% include "django/forms/widgets/attrs.html" %}
   />
-  <label role="label" class="crt-checkbox__label" {% if widget.attrs.id %} for="{{ widget.attrs.id }}"{% endif %}>{{ widget.label }}</label>
+  <label role="label" class="crt-checkbox__label" {% if widget.attrs.id %} for="{{ widget.attrs.id }}"{% endif %}>{% trans widget.label %}</label>
```
LindsayYoung, line 10: Labels should also be served with lazy loading, these should be covered.

crt_portal/cts_forms/templates/forms/report_data.html
Jkrzy: You're right, this one was confusing to track down....:

These translation tags were needed because the `HateCrimesandTrafficking` and `ProtectedClass` models aren't using the defined `choices` as expected. They're currently storing the  human-readable name, as lazy loaded when each object instance was created, as opposed to the value that we would expect. 

Example:
`Threatened, forced, and held...` is in the database instead of `physical_harm`. This breaks the `get_FOO_display` and default widget rendering as the value from the database isn't present in the choice list.

Visible in the admin as well within on the associated model and instance pages. 

It looks like this dates back to ec54a7645b2a8f364f78908149b31eff5b9a16d8 and impacted how some of the other templates were implemented, I'll start looking at how to remedy.

crt_portal/cts_forms/templates/forms/report_data.html
Jkrzy: Addressed in #437

crt_portal/cts_forms/templates/forms/widgets/usa_checkbox_option.html
Jkrzy: Addressed in #437

PR #434, opened 2020-04-30
updates for screen reader

LindsayYoung:
This usually gets the status updates read by voice over, still a bit flaky.
I can't seem to do any better- I tied to fix it a few different ways. 
[[a11y] Add an alert when the details page is updated#442](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/442)

## What does this change?
- Usually reads the status alerts
    - adds  screen reader only alerts for the updates after page load, so they **should** be read by screen readers. (I want to try this with Terri and see if there might be a voiceover bug)
- when the form is update it links to the #status, this helps people using screen readers because they don't need to tab all the way into the screen

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/views.py
+++ b/crt_portal/cts_forms/views.py
@@ -275,7 +275,8 @@ class ShowView(LoginRequiredMixin, View):
             form.save()
             form.update_activity_stream(request.user)
             messages.add_message(request, messages.SUCCESS, form.success_message())
-            return redirect(report.get_absolute_url())
+            url = report.get_absolute_url()
+            return redirect(f"{url}#status-update")
         else:
             output = serialize_data(report, request, id)
             # Add form with errors to context
@@ -283,7 +284,12 @@ class ShowView(LoginRequiredMixin, View):
                 output.update({'contact_form': form})
             else:
                 output['actions'] = form
-            messages.add_message(request, messages.ERROR, form.FAIL_MESSAGE)
+            try:
+                fail_message = form.FAIL_MESSAGE
+            except AttributeError:
+                fail_message = 'No updates applied'
+
+            messages.add_message(request, messages.ERROR, fail_message)
 
             return render(request, 'forms/complaint_view/show/index.html', output)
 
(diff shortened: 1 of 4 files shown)


LindsayYoung:
Much cleaner implementation- thanks for the suggestion @Jkrzy

crt_portal/cts_forms/views.py
```diff
@@ -275,15 +275,21 @@ def post(self, request, id):
             form.save()
             form.update_activity_stream(request.user)
             messages.add_message(request, messages.SUCCESS, form.success_message())
-            return redirect(report.get_absolute_url())
+            url = report.get_absolute_url()
+            return redirect(f"{url}#status-update")
```
Jkrzy, line 279: This works but feels a little wonky to me, providing an ID we know won't always exist on the page, when folks reload the page, bookmark, share links, navigate w/ history, etc. 

Given the change in approach, moving the messaging into the title and the prominent display of the message box at the top of the content it feels less necessary to introduce this additional piece. Would you be ok with dropping this piece of the implementation?

LindsayYoung: This code is executed only when there are edits to the page. Wether those edits are successful or not, they should produce a status. 

The div isn't conditional, it should always exist and work, the status if there is one, is populated conditionally inside it. 
https://github.com/usdoj-crt/crt-portal/pull/434/files#diff-ab31ffb528fab745c40100187155b01eR60 
Though, we could move the div to be conditional, if the id doesn't exist, the page will still load normally so I don't see a downside. 

The status message and the link to the status are solving different problems. Linking to the status part of the page doesn't give additional information for screen reader users, like the status does. 

What adding the link in the page does do, is makes the page navigation much easier for seen reader users because the user starts inside the page after making an update. Otherwise, the user has to keep re-navigating inside the page every time there is an update to the page, which we have seen from testing adds friction.

Jkrzy: Ahh misread on my part, you're totally right that div does _always_ exist! That context for screen reader users is super helpful 👍 👍

PR #437, opened 2020-05-01
Field tweaks for HateCrimesandTrafficking and ProtectedClass models

Jkrzy:
## What does this change?
**targeting `i8n` branch**

For each `HateCrimesandTrafficking` and `ProtectedClass` models, create a new field `value` to which we migrate all processing and display functionality. Deprecating `HateCrimesandTrafficking.hatecrimes_trafficking_option` and `ProtectedClass.protected_class`. These deprecated fields should be removed in a later issue.

Includes necessary migrations, model, view, form, and test updates.

### Why?
Proposed resolution for the issue noticed [here](https://github.com/usdoj-crt/crt-portal/pull/432/files#r417708083).

Moving towards instances and usage of these models/fields which conforms to Django norms for [Charfields w/ choices](https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices). This allows us to use all of the functionality Django provides out of the box for these fields/instances and simplifies, by removing previous workarounds to accommodate the old implementation.

### Author

### Reviewer

## Notes for reviewer:


Jkrzy:
This combined with #432 is a decent chunk of changes, attempting to facilitate review a bit by pulling out this functionally separate piece that supports translation but doesn't itself implement it. 

If you'd prefer, happy to merge this into `i8n` first and then request review. Thanks @LindsayYoung!!

LindsayYoung:
The code looks solid. I am getting errors when I try to migrate my database. I am going to look into it Monday my 5pm Friday 🧠 isn't up for the challenge.

LindsayYoung:
@Jkrzy OK, I figured out the issue.

In my database, dev and stage, (I didn't check prod) there is a period at the end of trafficking:
 - Threatened, forced, and held against your will for the purposes of performing work or commercial sex acts. This could include threats of physical harm, withholding promised wages, or being held under a false work contract.

There is not in the model choices. The other choices generally don't have periods, so I do think no periods is correct, but we need to account for it in the migration.

Do you think we could do a `__contains` for that look up so it works regardless of the period?

Jkrzy:
Ahh interesting thanks! Tweak to handle that specific scenario incoming, taking a slightly different approach since I'd like to iterate through all existing objects allowing us to fail and not apply changes when/if we hit situations just like this one.

LindsayYoung:
Sounds good! Feel free to go with a better implementation if you have one in mind

Jkrzy:
Ready for another look @LindsayYoung, thanks!!

crt_portal/cts_forms/forms.py
```diff
@@ -859,7 +858,7 @@ def __init__(self, *args, **kwargs):
             error_messages={'required': PROTECTED_CLASS_ERROR},
             required=False,
             label=PROTECTED_CLASS_QUESTION,
-            queryset=ProtectedClass.objects.filter(protected_class__in=PROTECTED_CLASS_CHOICES).order_by('form_order'),
+            queryset=ProtectedClass.objects.all().order_by('form_order'),
```
LindsayYoung, line 861: There are more objects in the DB that we don't want to display.

crt_portal/cts_forms/migrations/0067_populate_new_choice_fields.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def reverse(apps, schema_editor):
+        """Empty the values places in `value`"""
+        HateCrimesandTrafficking = apps.get_model('cts_forms', 'HateCrimesandTrafficking')
+        ProtectedClass = apps.get_model('cts_forms', 'ProtectedClass')
+        for hc in HateCrimesandTrafficking.objects.all():
+            hc.value = ''
+            hc.save
+
+        for pc in ProtectedClass.objects.all():
```
LindsayYoung, line 51: OK, one more thing. Can we loop though the protected class options instead of the existing class objects 

We used to aggregate the data in protected class differently. So there are going to be more protected classes than we have protected class option. For example, we used to have 

 - National origin (including ancestry, ethnicity, and language)

Then, that was split into two groups. 
 - National origin (including ancestry and ethnicity)
 - Language

This was our first test case on making sure the data could evolve. It is used as the example in the protected class changes in the infrequent tasks and maintenance docs: https://github.com/usdoj-crt/crt-portal/blob/develop/docs/maintenance_or_infrequent_tasks.md#change-protected-class-options.

I don't think this particular issue happens in Production since we only added data there recently, but that list is going to change over time and we want a way to preserve past data and update display when those things don't have one to one matches.

crt_portal/cts_forms/tests/tests.py
```diff
@@ -88,8 +74,8 @@ def test_Location_invalid(self):
         )
 
     def test_Class_valid(self):
-        form = ProtectedClassForm(data={
-            'protected_class': ProtectedClass.objects.filter(protected_class__in=PROTECTED_CLASS_CHOICES),
+        form = ProtectedClassForm({
+            'protected_class': ProtectedClass.objects.all(),
```
LindsayYoung, line 78: I don't think we will want all objects especially over time as this list changes and we need to preserve the previous data

Jkrzy:
Thanks for the notes @LindsayYoung! How the evolution of options for `ProtectedClass` and `HateCrimesandTrafficking` is currently handled is soooo much clearer now.

I've got an implementation...commits on the way

Jkrzy:
Took another pass at this one and ready for your 🧠 @LindsayYoung. We're still setting values for all instances since we want to make the new field unique.

The newer piece is a manager for `ProtectedClass` that allows us readily filter for what are considered "active" choices, those defined in `model_variables.py`

PR #441, opened 2020-05-01
[WIP] Hotfix/cors update

LindsayYoung:
The new domains are woking but the form is not working with the error:
`ERR Forbidden (Referer checking failed - no Referer.): /report/`

I think the next step is to add the new urls to the trusted origins. Want to double check the implications on this.


Jkrzy:
Ooo, I think we need to update out CDN service to pass along at least the `Referer` header: https://cloud.gov/docs/services/cdn-route/#more-about-how-the-cdn-works

LindsayYoung:
Updated the CDN service to pass on the referer. Referer whitelisting were not the issue. Thank for your help on this @Jkrzy

PR #443, opened 2020-05-05
label "required" fields

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/319)

## What does this change?

- Removes `(Optional)` wording
- Adds `required` wording (with an asterisk before it)

### DIFF
--- a/crt_portal/cts_forms/model_variables.py
+++ b/crt_portal/cts_forms/model_variables.py
@@ -308,7 +308,6 @@ DATE_ERRORS = {
     'no_past': _('Please enter a year after 1900.'),
     'not_valid': _('Please enter a valid date. Use format MM/DD/YYYY.'),
 }
-INCIDENT_DATE_HELPTEXT = _('You must enter a month and year. Please use the format MM/DD/YYYY.')
 
 VOTING_ERROR = _('Please select the type of election or voting activity.')
 
--- a/crt_portal/static/sass/custom/form.scss
+++ b/crt_portal/static/sass/custom/form.scss
@@ -364,3 +364,26 @@ select.usa-input--error {
     max-width: 80%;
   }
 }
+
+.field-required {
+  font-size: 1rem;
+  font-weight: normal;
+  font-style: italic;
+  color: $gray-50;
+
+  &::before {
+    content: "*";
+    font-weight: bold;
+    margin-left: 0.3rem;
+    color: $red-vivid-60;
+  }
+
+  &--group {
+    @extend .field-required;
+
+    &::before {
+      @extend .field-required::before;
+      margin-left: 0.5rem;
+    }
+  }
+}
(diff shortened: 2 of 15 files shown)


cantsin:
Side note, I'm not sure why `secondary-dark` is giving us a slightly different red color: it's supposed to be `red-60v`: #B51D09 but we're getting #B50909.

crt_portal/static/sass/custom/form.scss
```diff
@@ -359,3 +359,26 @@ select.usa-input--error {
...
+  font-size: 1rem;
+  font-weight: normal;
+  font-style: italic;
+  color: $gray-50;
+
+  &::before {
+    content: "*";
+    font-weight: bold;
+    margin-left: 0.3rem;
+    color: $red-vivid-60;
```
cantsin, line 373: I would have preferred `@include u-text('red-60v');` but this doesn't give us the expected #B51D09 value (we get #B50909 instead)

crt_portal/cts_forms/model_variables.py
```diff
@@ -308,7 +308,6 @@
     'no_past': _('Please enter a year after 1900.'),
     'not_valid': _('Please enter a valid date. Use format MM/DD/YYYY.'),
 }
-INCIDENT_DATE_HELPTEXT = _('You must enter a month and year. Please use the format MM/DD/YYYY.')
```
cantsin, line 311: duplicate line (see line 300)

LindsayYoung:
Went through the form and found 3 small things.

"Did this happen while in custody or incarcerated?" missing required- I think we need to add it here:
https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/templates/forms/question_cards/police_location.html#L3

"Please choose the type of location that best describes where the incident happened?" " missing required- I think we need to add it here:
https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/templates/forms/question_cards/commercial_public_location.html#L5

I also noticed that we need this to be on one line for the error message to render. (You didn't introduce this error.)
https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/templates/forms/question_cards/commercial_public_location.html#L22-L23

(For testing this, I went through the different primary issues, that is where all the branching logic in the form stems from.)

Also for future reference, would you prefer I push small fixes to your branch, or just flag them like this?

cantsin:
Thanks!! Great catches! Please feel free to push fixes to this branch if you are so inclined. Will squash and add Co-authored-by trailers when that happens.

cantsin:
Meanwhile I'm happy to fix these errors too, just let me know. I'll finish up one other PR then return to this.

LindsayYoung:
Cool, circling back to this now. I add the small bits I mentioned, do one more runthrough and merge if it goes well

Feel free to make small edits on my PRs. I am not particular about git histories. (Feel free to comment when my histories get a bit messy- its one of my flaws.)

cantsin:
I think we're good to go here! Went through all the different sections plus CRT form, can't see anything else we might be missing.

cantsin:
Rebased + added Co-authored-by.

PR #409, opened 2020-04-21
Assign a complaint to individual staff member

Jkrzy:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/209)

## What does this change?

Adds `Report.assigned_to` field which allows a `User` to be associated with a complaint. 
Assignee can be changed w/ `ComplaintAction` form.

Current implementation uses [accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete) to assist users w/ navigation of what is expected to be a large list of potential assignees. 

## TODO
- [x] Sort out styling of autocomplete input, could use some help here 😄 
- [ ] Confirm behavior of autocomplete 

## Screenshots (for front-end PR):
Screen Shot 2020-04-23 at 8 48 43 PM

### Author

+ [ ] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.)
+ [ ] Check for [accessibility](/docs/a11y_plan.md).
+ [ ] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

+ [ ] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.)
+ [ ] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

## Notes for reviewer:

See [PR instructions doc](https://github.com/usdoj/crt-portal/blob/master/docs/pull_requests.md) for full pull request review instructions.

PR #445, opened 2020-05-06
Small IE header fixes + click target expansion

hbillings:
Should close out [#409](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/409) and [#375](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/375), which were both mostly done already, but with the IE fixes from yesterday, there were a few outstanding small issues.

## What does this change?
- Changes the way position of the progress bar is calculated in IE, so that the progress bar will always "un-stick" at the right point (a bug in the current method meant it would sometimes cover up the other part of the header)
- Adds some CSS black magic to the proform card headers so that IE users can still see the card header when they jump to it. (previously these were covered by the static progress bar, which was confusing)
- BONUS: Changes the click targets in the progress bar to include the number (this took 5 mins and I was in the file anyway)


crt_portal/cts_forms/templates/forms/pro_template.html
```diff
@@ -8,7 +8,7 @@
 {% endblock %}
 
 {% block form_questions %}
-<div class="crt-portal-card">
+<div class="crt-portal-card crt-portal-card--proform">
```
hbillings, line 11: These extra classes are just so I can target only the proform  `h2`s with CSS. Otherwise the hack for getting the headings to show would get applied to every card in the app, which seemed unnecessary.

crt_portal/cts_forms/templates/forms/pro_template.html
```diff
@@ -23,16 +23,17 @@ <h2 id="intake">Intake</h2>
...
   <div class="crt-portal-card__content crt-portal-card__content--lg">
-    <a name="contact"></a>
     {% include "forms/question_cards/contact.html" with card_title='Contact' %}
   </div>
 </div>
 
-{% include "forms/question_cards/single_form.html" with field=form.servicemember card_title="Service member" %}
+<div class="crt-portal-card--proform">
```
hbillings, line 32: Had to wrap those extra classes around the Service Member and Incident Location cards, since the markup for those was entirely contained in their components.

crt_portal/cts_forms/templates/forms/pro_template.html
```diff
@@ -23,16 +23,17 @@ <h2 id="intake">Intake</h2>
   </div>
 </div>
 
-<div class="crt-portal-card">
+<div class="crt-portal-card crt-portal-card--proform">
   <div class="crt-portal-card__content crt-portal-card__content--lg">
-    <a name="contact"></a>
```
hbillings, line 28: `name` is deprecated, and all of the others were getting targeted via ID already, so just changed this to match!

crt_portal/static/js/progress-bar.js
```diff
@@ -15,8 +15,7 @@
     } catch (err) {
       console.log(err);
     }
-    // return supportsScroll;
-    return false; // JUST FOR TESTING, DUDE
```
hbillings, line 19: oops

crt_portal/static/js/progress-bar.js
```diff
@@ -43,8 +42,9 @@
     }
   } else {
     // if browser doesn't support scrolling and position: sticky, use position: fixed instead
+    var header = document.getElementsByClassName('intake-header')[0];
+    var barOffset = header.getBoundingClientRect().height;
     var bar = document.getElementsByClassName('intake-header--progress-bar')[0];
-    var barOffset = bar.getBoundingClientRect().top;
```
hbillings, line 47: if the page was loaded to an anchor tag, this would equal 0, which would have the effect of permanently sticking the progress bar to the top of the window. oops #2

crt_portal/static/sass/custom/intake.scss
```diff
@@ -57,6 +57,15 @@
...
 
+// Fix for IE and Edge, which rely on .intake-header--progress-bar being position: fixed (applied via progress-bar.js).
+// This prevents the progress bar from covering up the card header when a target is clicked on.
+.crt-portal-card--proform h2::before {
+  display: block;
+  content: '';
+  height: 150px;
+  margin: -150px 0 0;
+}
+
```
hbillings, line 68: don't ask me why this works I don't have any answers

crt_portal/static/sass/custom/progress.scss
```diff
@@ -64,17 +64,21 @@ ol.steps {
...
+
+    a {
+      display: block;
+    }
   }
 
-  .step::before,
-  .step::after {
+  .step a::before,
+  .step a::after {
```
hbillings, line 74: changing these rules to target the `a` instead of the `li.step` was all it took to make the numbers clickable also!

hbillings:
@LindsayYoung I'm not entirely sure what a `.mo` file is -- seems to be related to translation? It didn't exist in `develop` but for some reason did exist in my local branch, so I removed it to match `develop`. 🤷

LindsayYoung:
No worries!  You are right aobut the .mo and we just added this to the gitignore
https://github.com/usdoj-crt/crt-portal/blob/develop/.gitignore#L20

crt_portal/static/sass/custom/progress.scss
LindsayYoung: Nice!!!

LindsayYoung:
It seems like there is more space between the header and the section heading than there was before. (Looking on Chrome)

This is how it lines up if I click on primary concern:
Screen Shot 2020-05-06 at 4 09 43 PM

hbillings:
@LindsayYoung oooh, good catch! I just pushed an updated version that only applies that space to the IE/Edge version of the header that's using `position: fixed`.

PR #447, opened 2020-05-06
Fix dev regressions

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/473)

## What does this change?

Fixes dev regressions with: button colors, form fieldset spacing, and removes a stray closing template tag.

### DIFF
--- a/crt_portal/static/sass/custom/form.scss
+++ b/crt_portal/static/sass/custom/form.scss
@@ -95,6 +95,10 @@ form#report-form {
       margin-top: 1.5rem;
     }
 
+    &:last-of-type:not(:only-of-type) {
+      margin-top: 2.5rem;
+    }
+
     ul {
       margin-bottom: 0;
     }
--- a/crt_portal/static/sass/custom/form.scss
+++ b/crt_portal/static/sass/custom/form.scss
@@ -115,6 +115,7 @@ form#report-form {
   }
 
   .light-button {
+    background-color: transparent;
     &:hover {
       background-color: $white;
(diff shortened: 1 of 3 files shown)


cantsin:
@hbillings turns out that some of the above issues came in with 8872dc3a90afd063d2f3563400dd5c6bcd365b8a 

- for the buttons

```scss
button,
.usa-button {
  background-color: $blue-warm-vivid-80;
}
```

- for the spacing

```scss
    &:last-of-type:not(:only-of-type) {
      margin-top: 2.5rem;
    }
```

I realize we may not want to bring back some of these changes. I'm missing some prior context and history -- specifically, I would prefer to remove .usa-button background-color entirely but I'm not sure why that change was introduced. Please let me know what I should do here!

LindsayYoung:
I am happy to review this one, since @hbillings has limited time.

crt_portal/static/sass/custom/form.scss
```diff
@@ -95,6 +95,10 @@ form#report-form {
       margin-top: 1.5rem;
     }
 
+    &:last-of-type:not(:only-of-type) {
```
hbillings, line 98: I should probably throw a comment in here explaining what this does. I'll do that as part of the rest of the design audit finishes. (I'd add it as a suggestion now, but I can't actually remember precisely what it does, ha)

PR #448, opened 2020-05-06
Complaint record styling

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/426)

## What does this change?

On crt form: adjust card spacing to be ~2.5 rem all around~ 1.5em/24px between stacked cards;  32px gutter; de emphasize primary complaint text

### DIFF
--- a/crt_portal/cts_forms/templates/forms/complaint_view/show/index.html
+++ b/crt_portal/cts_forms/templates/forms/complaint_view/show/index.html
@@ -61,7 +61,7 @@
         {% include 'partials/messages.html' %}
       </div>
     </div>
-    <div class="grid-row grid-gap-lg">
+    <div class="grid-row grid-gap-5">
       <div class="tablet:grid-col-4 grid-offset-1">
         {% include 'forms/complaint_view/show/actions.html' with title="Actions" icon="img/intake-icons/ic_check-circle.svg" %}
         <div class="activity-stream">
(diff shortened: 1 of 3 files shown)


crt_portal/cts_forms/templates/forms/complaint_view/show/index.html
```diff
@@ -61,7 +61,7 @@ <h2 class="margin-right-205 margin-top-0 margin-bottom-0 backend-blue">
         {% include 'partials/messages.html' %}
       </div>
     </div>
-    <div class="grid-row grid-gap-lg">
+    <div class="grid-row grid-gap-5">
```
cantsin, line 64: Not sure if we wanted to change the gutter size here, but did so for consistency's sake. Easy to undo if we want to keep as-is.

cantsin:
I think this is a quick and easy ticket to resolve but please do let me know if I'm missing something else here :sweat_smile:

LindsayYoung:
I just wanted clarification on the original ticket- @Jacklynn or @AvivaOskow Do we want 1.5 rem on all cards public view and crt view or only crt views?

This is the current front end card:
Screen Shot 2020-05-06 at 6 23 26 PM

This would be the 1.5 rem front end card spacing:
Screen Shot 2020-05-06 at 6 23 58 PM

This is the current backend card spacing:
Screen Shot 2020-05-06 at 6 21 24 PM

This would be the 1.5 rem backend card spacing:
Screen Shot 2020-05-06 at 6 22 40 PM

cantsin:
Changed -- as per feedback we want 24px between both public/CRT cards.

PR #458, opened 2020-05-11
New footer

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/318)

## What does this change?

Adds layout for footer. (WIP: still needs styling)


LindsayYoung:
Great start! Thanks for the quick turnaround.

We need to remove the links that won't be built this sprint:
 - Your Rights
 - Already submitted?

We can add those back when they are built.

We are going to put the privacy policy on the CRT website, that can stay and we can update that once we have a url for it. (Noting that It was also requested that that open in a new tab.)

PR #467, opened 2020-05-14
Fix owasp scan (dev)

LindsayYoung:
https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/504

- Removing blank=true to reduce risk of [Reverse Tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing) 
- Exempt the language cookie from evaluation, that only got rid of one exception
   - Cookie Without Secure Flag 

For now the cookie specific exemption isn't working for these so we are ignoring:
   - Cookie Poisoning
   - Cookie Without SameSite Attribute

Confirmed with @Jkrzy that this is just the language cookie and wouldn't have security implications.


.circleci/config.yml
```diff
@@ -79,7 +79,7 @@ jobs:
             name: Pull owasp zap docker image and run scan
             command: |
               docker run -v $(pwd):/zap/wrk/:rw -t owasp/zap2docker-weekly zap-baseline.py \
-              -t https://crt-portal-django-dev.app.cloud.gov/report/ -c .circleci/zap.conf
+              -t https://crt-portal-django-dev.app.cloud.gov/report/ -c .circleci/zap.conf -z "-config rules.cookie.ignorelist=django_language,_ga"
```
LindsayYoung, line 82: I am testing out the code here and will update the other places accordingly

PR #499, opened 2020-05-26
Adding instructions for upgrading databases

LindsayYoung:
Update to docs with example of how to upgrade the databases


ilodidoj:
These instructions are straight forward. No change needed.

PR #522, opened 2020-06-02
Clean up color tokens and variables

hbillings:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/525) - note that this is only a partial solution!

## What does this change?
There were two separate files where Sass variables were being declared. I got rid of `variables.scss` in favor of using the USWDS standard child theme files. Then, I combed through the CSS and HTML to enforce consistent usage of the USWDS color tokens. 

I also wrote a couple of new classes for the privacy footer so that I could move those styles out of the HTML and into the Sass, where I could use color variables for them.

Note to @avivia-oskow and @Jacklynn: there were a couple of instances where a color variable was only used once and was very, very similar to an existing color that saw more widespread usage. In those cases, I changed the less-used color to the more-used color. (Think `gray-5` replacing `gray-cool-5` in a table border on the admin view.) I'd really appreciate it if y'all could take a spin through this on dev (if someone can tell me how to get this up there) and let me know if anything seems off.

[hopefully this is a no-op change!]

### Author

### Reviewer

## Notes for reviewer:


crt_portal/static/sass/_uswds-theme-color.scss
```diff
@@ -24,53 +24,53 @@ Theme palette colors
...
 $theme-color-base-ink:              'gray-warm-80';
 
 // Primary colors
 $theme-color-primary-family:        'blue';
-$theme-color-primary-lightest:      false;
-$theme-color-primary-lighter:       'blue-10';
-$theme-color-primary-light:         'blue-30';
+$theme-color-primary-lightest:      'blue-warm-5';
+$theme-color-primary-lighter:       'blue-warm-10';
+$theme-color-primary-light:         false; // 'blue-30'
```
hbillings, line 41: Theme colors that we're not using anywhere are set to `false`. This shouldn't affect anything, but I thought it might make things clearer for future front-end devs trying to figure out which colors to use.

crt_portal/static/sass/_uswds-theme-color.scss
```diff
@@ -24,53 +24,53 @@ Theme palette colors
...
 $theme-color-base-lightest:         'gray-5';
-$theme-color-base-lighter:          'gray-cool-10';
-$theme-color-base-light:            'gray-cool-30';
-$theme-color-base:                  'gray-cool-50';
+$theme-color-base-lighter:          'gray-warm-10'; // gray-cool-10
+$theme-color-base-light:            'gray-30'; // gray-cool-30
+$theme-color-base:                  'gray-50'; // gray-cool-50
 $theme-color-base-dark:             'gray-cool-60';
-$theme-color-base-darker:           'gray-cool-70';
+$theme-color-base-darker:           'gray-70'; // gray-cool-70
```
hbillings, line 33: We really don't use the cool grays anywhere (although it seems like some of the USWDS base elements use `$theme-color-base-dark` -- maybe that should be updated to a warm gray also?)

crt_portal/static/sass/custom/alerts.scss
```diff
@@ -2,13 +2,14 @@
 
 .page-errors {
   font-weight: bold;
-  color: $red-vivid-60;
+  color: color($theme-color-secondary-dark);
```
hbillings, line 5: I was a little torn on how to do this. On one hand, using the USWDS `color()` function in the CSS rules themselves isn't going to be intuitive to new front-end devs. But on the other, the alternative is to use the `color()` function in the child theme colors file (i.e., `$theme-color-primary: color('blue-warm-70v')`), which then means you can't use the color variable in USWDS mixins (like `@u-bg($theme-color-primary)`).

Jacklynn:
I don't see the color styles you're removing in any of my annotations! 👍

crt_portal/static/sass/_uswds-theme-color.scss
cantsin: :thinking: Can we remove these `gray-...` comments and add an explanation as to why we are still keeping `gray-cool-60` for now?

crt_portal/static/sass/_uswds-theme-color.scss
cantsin: Great idea!

crt_portal/static/sass/custom/alerts.scss
cantsin: I think what you have is the best compromise. Agree doing `color()` is not the most intuitive but it's definitely good enough to move forward with.

crt_portal/static/sass/_uswds-theme-color.scss
hbillings: good idea! I'll add that, thanks!

PR #525, opened 2020-06-04
Convert typography to USWDS child theme

hbillings:

## What does this change?
**NOTE** This is going to be way easier to review if #522 gets reviewed and merged in first! I based this work on that PR.

- defines standard type scale based on mockups
- standardizes sizing and class name conventions for headers
- standardizes sizing for body type variations
- retrofits existing templates and Sass files with new names and rules, removing all hardcoded `rem` values in favor of mixins and variable names

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/templates/forms/confirmation.html
```diff
@@ -3,15 +3,15 @@
 {% load static %}
 
 {% block page_header %}
-<header class="confirmation-header">
```
hbillings, line 6: There were a few duplicative classes like this that implemented existing type rules under different class names. I simplified where it seemed to make sense.

crt_portal/static/sass/_uswds-theme-color.scss
```diff
@@ -24,57 +24,57 @@ Theme palette colors
 */
 
 // Base colors
-$theme-color-base-family:           'gray-cool';
```
hbillings, line 27: This file shouldn't be in here.  I think it's showing up because I based this on the previous color palette work.

crt_portal/static/sass/_uswds-theme-typography.scss
```diff
@@ -398,16 +398,20 @@ $theme-body-line-height:           5;
...
 
-// Headings
-$theme-h1-font-size:               '2xl';
-$theme-h2-font-size:               'xl';
-$theme-h3-font-size:               'lg';
+// Headings -- not used
+// NOTE: We don't use these because the design system's font token normalization 
+// changes the size of the font on the fly:
+// https://designsystem.digital.gov/design-tokens/typesetting/overview/
+// Instead, we use REM defined in the next section.
```
hbillings, line 405: I HAVE FEELS ABOUT THIS

crt_portal/static/sass/_uswds-theme-typography.scss
```diff
@@ -420,3 +424,69 @@ $theme-lead-font-family:           'heading';
...
+  }
+}
+
+@mixin text--heading__xlarge {
+  font-size: $theme-heading-font-size-xl;
+  line-height: 1.3;
+  @include at-media-max(mobile-lg) {
+    font-size: $theme-heading-font-size-lg;
+  }
+}
```
hbillings, line 492: I don't entirely love putting custom code in the `_uswds-theme` files, but those are also the first files I'd look at if I were trying to figure out what was going on. I'm open to putting these somewhere else if that makes sense to other folks

cantsin:
Its a bit hard for me to review, can we compare this branch against `hjb/uswds-child-theme` instead?

crt_portal/static/sass/_uswds-theme-typography.scss
cantsin: SO MANY FEELS. 

At the risk of being labeled a heretic and cast into a lake of fire, I wonder if it wouldn't be easier to leave the default in for the sake of long-term maintenance. Eventually someone is going to accidentally use, say, the built-in USWDS `h1` mixin instead and fonts will be subtly different as a result. It is difficult to figure out where all of these `theme-*-font-size` variables are used by USWDS, so I don't know the full implications of this change.

crt_portal/static/sass/_uswds-theme-typography.scss
cantsin: I think this is a good place :+1:

hbillings:
We chatted about this in co-design and decided that the way to move forward here is to merge this as-is, since it's effectively a typography reset, and open an issue with USWDS to see if we can get them to add a flag to the codebase that turns off font normalization!

PR #528, opened 2020-06-05
[WIP] Stage sso test

LindsayYoung:
deploys SSO to staging as per docs: https://github.com/usdoj-crt/crt-portal/blob/develop/docs/maintenance_or_infrequent_tasks.md#code-changes


cantsin:
@Jkrzy FYI -- updating staging to have SSO so that John can re-test

PR #551, opened 2020-06-16
admin: clean up wording

cantsin:
[ no zenhub issue ]

## What does this change?

Cleans up admin wording -- specifically verbose names. 

### Author

### Reviewer

## Notes for reviewer:


cantsin:
@Jkrzy I'm not sure what to do here. Superusers already can access the admin and all models within it, as you know, so all I did here was to clean up the wording on the admin a bit.

The only thing I can think of is to assign explicit permissions for adding/editing/deleting response templates for these super users (again... not really that useful). I suppose if we did not have super users, we could add a data migration for permissions for specific groups. Open to ideas here.

Jkrzy:
Not much else to do here if we go down the path of managing these via the admin interface. Before committing to that, we should chat with the team about if that's the best path. I'm leaning back towards _no_ and having the templates tracked in the repository.

My comment from slack a few days ago:
>Right now, all of the models besides Report are updated with code changes and migrations. This slows us down but gives us consistency across environments and a more robust way to test changes to form questions, protected classes, and these templates.
>
>Moving the changes to the Admin let’s us move a faster (though a hotfix is pretty quick as it is) and make changes directly in the environment of our choosing but requires manual entry across environments and bypasses the existing code review, merge, and testing process.

cantsin:
@Jkrzy i would like to get this in if nothing else to clear up the PR board -- this just cleans up the naming on admin models. (no zenhub issue involved here anymore)

If we don't want this in, I'll close; doesn't matter to me. Either way, let's get rid of this dangling PR.

PR #580, opened 2020-07-10
data migrations to remove assignee for old closed reports 

harikrishnagit:

## What does this change?
For closed reports prior to 6/30/2020, their assignee value is to None

### Author

'migrate' ran successfully.  

### Reviewer

## Notes for reviewer:
Need help to load data to test this data migration. 


harikrishnagit:
@cantsin I'll rename the migration file. 
Do you use migration files to load data?, Or, do you load directly on local postgres using sql statements?
Also, I can use help with adding to activity log. When would you have time with this?

harikrishnagit:
> @harikrishnagit some things to note:
> 
>     1. please copy the zenhub issue in the description (`link-goes-here`) so it's easier to find.
> 
>     2. please rename the migration file to something more descriptive: it's hard to tell what the migration is doing from the filename alone.
> 
>     3. can we please have tests for this feature? this is a big change that will be done on production data and we need to make sure it works correctly.
> 
> 
> Regarding tests, we can do this by creating four test reports: two (one with assignee, one without) before 6/30 and another two (one with assignee, one without) after 6/30.
> 
> Also, and this one may be on me, now that I'm looking at this, we might want to specially call out this automated migration in the activity log. Joe alluded to this in the original ticket in a comment. Happy to discuss more when you have time.

# 1 and 2 are done. Would like to discuss #3, on how to do. 
Fixed the migration sequence breakage issue, and am ready for a demo with test data.

harikrishnagit:
> > @harikrishnagit some things to note:
> > ```
> > 1. please copy the zenhub issue in the description (`link-goes-here`) so it's easier to find.
> > 
> > 2. please rename the migration file to something more descriptive: it's hard to tell what the migration is doing from the filename alone.
> > 
> > 3. can we please have tests for this feature? this is a big change that will be done on production data and we need to make sure it works correctly.
> > ```
> > 
> > 
> > Regarding tests, we can do this by creating four test reports: two (one with assignee, one without) before 6/30 and another two (one with assignee, one without) after 6/30.
> > Also, and this one may be on me, now that I'm looking at this, we might want to specially call out this automated migration in the activity log. Joe alluded to this in the original ticket in a comment. Happy to discuss more when you have time.
> 
> # 1 and 2 are done. Would like to discuss #3, on how to do.
> 
> Fixed the migration sequence breakage issue, and am ready for a demo with test data.

 @cantsin , I paired with @ilodidoj earlier for data migration, which seem to work fine. We had a discussion about the activity log, which we need some direction from, like, 
- the data structure for activity log 
- which user should we use 
- use UTC to display or eastern time. 

Lmk if a call to discuss this helps, or co-ordinate here.

harikrishnagit:
> Thanks!
> 
> Looking at this I see two migration files: `remove-aassignee-iss613-migration.py` and `remove_assignee_20200714.py`. Do we want to run both of these? If so, can we merge them in one migration if possible?

We don't need either one of them. Should we remove these? 

> 
> Side note, please start the migration with a number so its easier to know which order they should be applied with. I believe the last one is `0084_form_template_standardization.py` so we should be starting with `0085_` here.

cantsin:
> We don't need either one of them. Should we remove these?

If we're not using either migration (I see you added another migration file) then absolutely, let's remove these two migration files. Side note, you can do `git mv` to rename a file if necessary (git should pick the rename up automatically tho).

harikrishnagit:
@cantsin , fixed the no user on circleci issue. you were thinking of another approach, which probably is more elegant. lmk your thoughts.

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,48 @@
...
+        # get all reports
+        reports = apps.get_model('cts_forms', 'Report')
+        # user = apps.get_model('cts_forms', 'User')
+        User = get_user_model()
+        userlist = User.objects.all().order_by('id')
+        # // Assuming that first user is the root user or superuser. If there are on users, create a dummy one
+        if len(userlist) > 0:
+            superuser = userlist[0]
+        else:
+            superuser = User.objects.create_user('migration_user')
```
cantsin, line 27: This isn't a bad idea, but my recommendation would be to skip the migration if we don't have an user. I think this keeps things somewhat simpler.

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # superuser = User.objects.get(pk=1)
+        # Register Report to be streamed
+        registry.register(reports)
+        timezone = pytz.timezone("UTC")
+        july1 = timezone.localize(datetime(2020, 7, 1))
+        for report in reports.objects.all():
+            # remove assignee for report closed date is before june 30 2020,
+            if report.closed_date is not None:
+                print(report.closed_date)
```
cantsin, line 37: remove `print` -- not needed here

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,48 @@
...
+                if report.closed_date < july1:
+                    report.assigned_to = None
+                    report.save()
+                    # Add the closed activity to activity stream
+                    action.send(superuser, verb='Assignee Removed: ',
+                                description='Removed assignee for closed record before July 1, 2020',
+                                target=report)
```
cantsin, line 44: :raised_hands:

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,48 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cts_forms', '0084_form_template_standardization'),
+    ]
+
+    def remove_assignee(apps, schema_editor):
+        # get all reports
+        reports = apps.get_model('cts_forms', 'Report')
+        # user = apps.get_model('cts_forms', 'User')
```
cantsin, line 20: I like the usage of `get_user_model` here -- can we remove this comment?

crt_portal/cts_forms/fixtures/assignee-removed-report.json
```diff
@@ -0,0 +1,282 @@
+[
```
cantsin, line 1: Thank you for this fixture! Would appreciate some better fake data though, `"violation_summary": "dfdafsadfasd"` isn't too representative. `"violation_summary": "violation summary here"` would be sufficient.

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
harikrishnagit: Yes, i'll remove commented out code.

harikrishnagit:
@cantsin question, at this point, after all reviews, do I/author merge the PR, or do you/reviewer?

cantsin:
The reviewer/approver merges in. Looking at this again now! :eyes:

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,42 @@
...
+            # get all reports
+            reports = apps.get_model('cts_forms', 'Report')
+            # Register Report to be streamed
+            registry.register(reports)
+            timezone = pytz.timezone("UTC")
+            july1 = timezone.localize(datetime(2020, 7, 1))
+            for report in reports.objects.all():
+                # remove assignee for report closed date is before june 30 2020,
+                if report.closed_date is not None:
+                    print(report.closed_date)
```
cantsin, line 32: Remove print please?

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,42 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cts_forms', '0084_form_template_standardization'),
+    ]
+
+    def remove_assignee(apps, schema_editor):
+        User = get_user_model()
+        userlist = User.objects.all().order_by('id')
+        # // Assuming that first user is the root user or superuser. If there are on users, create a dummy one
```
cantsin, line 20: The last part of the comment doesn't apply now, remove?

crt_portal/cts_forms/fixtures/assignee-removed-report.json
cantsin: Thank you for updating `violation_summary` but I meant it as an example. The data here is still a bit noisy, in terms of having `asdf` in it, can we make it a bit nicer?

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
harikrishnagit: yes, or reword. just did that.

harikrishnagit:
@cantsin fixture script to load data is modified. please review,

docker-compose.yml
```diff
@@ -1,8 +1,10 @@
-version: '3.7'
+version: "3.7"
 
 services:
   db:
     image: postgres:10.1-alpine
+    ports:
+      - 5432:5432
```
cantsin, line 7: Can we please remove this change? If you have a database running, this port forwarding setting conflicts and crashes the docker-compose invocation.

```
ERROR: for db  Cannot start service db: driver failed programming external connectivity on endpoint crt-portal_db_1 (2a0524dff68b3d8a0b69886e70b8090684d6876597db2a4ec90bc4a936ee06f0): Error starting userland proxy: listen tcp 0.0.0.0:5432: bind: address already in use
```

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
cantsin: @harikrishnagit :point_up: please remove the `print`?

crt_portal/cts_forms/migrations/0085_remove-assignee_20200716.py
```diff
@@ -0,0 +1,42 @@
...
+                    print(report.closed_date)
+                    if report.closed_date < july1:
+                        report.assigned_to = None
+                        report.save()
+                        # Add the closed activity to activity stream
+                        action.send(superuser, verb='Assignee Removed: ',
+                                    description='Removed assignee for closed record before July 1, 2020',
+                                    target=report)
+    operations = [
+        migrations.RunPython(remove_assignee),
```
cantsin, line 41: Do we want this migration to be reversed? Even if not, I'd still suggest throwing in a dummy reverse migration so we can roll back when needed (easier for local dev imo).

harikrishnagit: @cantsin, can we use reverting migration by running previous migrations?. Read that this is one option, don't know if you have any good/bad experience?

PR #611, opened 2020-08-07
improve pa11y failures

cantsin:
Partially addresses [Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/589)

## What does this change?

Tweaks pa11y to hopefully not fail so often. (I'm sure it will still fail, at this point I want to reduce the frequency)

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/pa11y_scripts/.crt-view.json
+++ b/pa11y_scripts/.crt-view.json
@@ -8,10 +8,10 @@
             "set field #id_username to pa11y_tester",
             "set field #id_password to imposing40Karl5monomial",
             "click #login",
-            "wait for url to be http://127.0.0.1:8000/form/view/",
-            "wait for element #id_assigned_to to be visible",
+            "wait for url to be http://127.0.0.1:8000/form/view/?status=new&status=open&no_status=false",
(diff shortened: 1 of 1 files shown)


cantsin:
I'll go ahead and push this change in, just so we can finally get `develop` to deploy properly.

Jkrzy:
Just noting, this won't resolve any intermittent timeouts -- just those in which the test script was already incorrect due to the `url` being waited for never arising likely because we're specifying URLParams.  We can use `path` here instead to be a bit more flexible. https://github.com/pa11y/pa11y#wait-for-fragmentpathurl

cantsin:
Yes. I tested with `path` but oddly enough it broke more often than not. I have to investigate why. I really wish there was a way to log which action pa11y failed at.

harikrishnagit:
same here, it is breaking more often, and at times it doesn't recognize elements which it does sometimes and which are clearly there. At one time, the `path` worked for me, it doesn't  in last few runs.

PR #599, opened 2020-07-22
Issue608

MattBowles8:
[Issue 608](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/608)

### Author

+ [x] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.)
+ [ ] Check for [accessibility](/docs/a11y_plan.md).
+ [ ] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

+ [x] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.)
+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

## Notes for reviewer:

See [PR instructions doc](https://github.com/usdoj/crt-portal/blob/master/docs/pull_requests.md) for full pull request review instructions.

PR #616, opened 2020-08-10
Remove extra padding above question headers

hbillings:
Hotfix.

## What does this change?
[A recent commit removed an inline style in favor of adding it to the class that existed on the heading](https://github.com/usdoj-crt/crt-portal/commit/1d4129cf857874ab51c9254b817d8f31b77d5c05). This created a visual regression because that padding does not need to be applied to every `question--header` element (and in fact probably should not even have been in service on the original single element).

Screen Shot 2020-08-10 at 5 52 49 PM

Screen Shot 2020-08-10 at 5 53 01 PM

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/static/sass/custom/form.scss
+++ b/crt_portal/static/sass/custom/form.scss
@@ -8,20 +8,12 @@ form#report-form,
 form#filters-form {
   .question--header {
     width: 100%; // IE11 fix
-    // margin-top: 0.5rem;
-    // margin-bottom: units($spacer-3x);
-    padding-top: 32px;
     margin: 0;
-    + .help-text,
-    + .help-text__small,
-    + p {
-      // margin-top: -1rem; // to match other help text margins
-    }
   }
 
   .zip--code--margin {
     margin-top: 0;
-    @include at-media-max(mobile-lg){
+    @include at-media-max(mobile-lg) {
       margin-top: 1.5rem;
     }
   }
@@ -280,7 +272,7 @@ form#filters-form {
         margin-bottom: units($spacer-1x);
         padding-left: units($spacer-5x);
         border-bottom: 1px solid color($theme-color-base-lighter);
-        padding-bottom: .5rem;
+        padding-bottom: 0.5rem;
       }
 
       ul {


crt_portal/static/sass/custom/form.scss
```diff
@@ -8,20 +8,12 @@ form#report-form,
...
   .question--header {
     width: 100%; // IE11 fix
-    // margin-top: 0.5rem;
-    // margin-bottom: units($spacer-3x);
-    padding-top: 32px;
     margin: 0;
-    + .help-text,
-    + .help-text__small,
-    + p {
-      // margin-top: -1rem; // to match other help text margins
```
hbillings, line 18: This shit was stuff I left in when I was doing the spacing refactor, and should have been removed ages ago. My bad!

crt_portal/static/sass/custom/form.scss
```diff
@@ -8,20 +8,12 @@ form#report-form,
...
-    }
   }
 
   .zip--code--margin {
     margin-top: 0;
-    @include at-media-max(mobile-lg){
```
hbillings, line 24: linting fix

crt_portal/static/sass/custom/form.scss
```diff
@@ -280,7 +272,7 @@ form#filters-form {
         margin-bottom: units($spacer-1x);
         padding-left: units($spacer-5x);
         border-bottom: 1px solid color($theme-color-base-lighter);
-        padding-bottom: .5rem;
```
hbillings, line 283: linting fix

crt_portal/static/sass/custom/form.scss
```diff
@@ -8,20 +8,12 @@ form#report-form,
 form#filters-form {
   .question--header {
     width: 100%; // IE11 fix
-    // margin-top: 0.5rem;
-    // margin-bottom: units($spacer-3x);
-    padding-top: 32px;
```
cantsin, line 13: :thinking: do we know what this padding was originally a fix for?

Jacklynn:
partyparrot

crt_portal/static/sass/custom/form.scss
hbillings: @cantsin [The commit that introduced it](https://github.com/usdoj-crt/crt-portal/pull/599/commits/636e0685af8da1d531382de4eefe36c98949f4ac) was also in #559, just called "completed with 8"? I tracked it back to the original issue, where "FORM - Personal Characteristics: Spacing between header and question" was listed as the eighth checkbox, with this screenshot:


 The spacing issue between "Personal Characteristics" and the question header is something I've never been able to reproduce locally or have seen on prod.  After removing the padding added in that commit, everything also looked correct to me. 🤷 I'd be really curious how it looks to other folks now!

Regardless, I'm concerned about the approach taken. If extra padding _was_ needed here, the right approach would have been to use the USWDS `padding-top-4` classname inline, not to either add an inline style or (especially) to add it to a classname that's used widely across the app.

cantsin: Thanks! I just double checked, `Personal Characteristics` spacing seems good to me too. I agree with the `padding-top-4` approach too, I believe we missed this during review :disappointed: 

Thank you so much for your help with this! :bow:

PR #634, opened 2020-08-19
Re-introduce pa11y-ci tests to CI

Jkrzy:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/589)

## What does this change?
We re-introduce and limit the scope of our pa11y-ci tests to use a single instance of pa11y-ci and evaluate all pages not requiring progression through the report submission form.

Using a single pa11y-ci instance here avoids having to serve all static assets for each url examined.

This gives us pa11y coverage on the landing page, first step of /report/ form, and the privacy-policy page.

### Author

### Reviewer

## Notes for reviewer:


pa11y_scripts/.without-report
```diff
@@ -0,0 +1,24 @@
+{
+    "urls": [{
+            "url": "http://127.0.0.1:8000/",
```
cantsin, line 3: Do we know if the url is fully loaded after this point? Should we wait for an element on this page to be visible?

Jkrzy: pa11y waits for headless chrome to open and load the provided URL before running. We only need a wait action if we're interacting with the page, navigating to a new page, or there's some other kind of page modifications _after_ the page has been loaded.

cantsin: Thanks, works for me.

PR #635, opened 2020-08-19
Complaint detail navigation

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/666)

## What does this change?

On the detail page, it is now possible to navigate through the currently filtered reports. Note that modifying the current report may eliminate the report itself from the current filter -- in that case, we mark the current index as N/A but still allow the user to navigate backwards or forwards.

### Author

### Reviewer

## Notes for reviewer:


hakhalid11:
@cantsin When testing this I found that clicking apply changes with out any fields change results in not preserving the previous/next index.

image

the screenshot above is the result of just hitting apply changes button without any change. Also the url loses all the request params so this leads me to believe that maybe the fix should be in views.py.

cantsin:
@hakhalid11 great catch. turns out that the `next` query parameter was not being preserved in that specific use case -- submit empty data -- i believe this was an oversight from the previous implementation. Fixed.

cantsin:
@Jkrzy can you elaborate a bit? I can't replicate -- if I edit violation summary or the report form, everything works as advertised. Can you give me some more detail on how to replicate?

Jkrzy:
Ah, the scope is more specifically limited to:

* Start with a search that limits the # of reports returned, like "Show me all reports with `location_city_town=TownA`
* View report, modify `location_city_town` to not be responsive to the 
* POST succeeds
* subsequent redirect includes `index=` which fails

crt_portal/cts_forms/views.py
```diff
@@ -287,6 +314,41 @@ def get(self, request, id):
...
+            previous_id = requested_ids[index - 1] if index > 0 else None
+            next_id = requested_ids[index + 1] if index < len(requested_ids) - 1 else None
+            next_query = urllib.parse.quote(return_url_args)
+            output.update({
+                'filter_count': report_query.count(),
+                'filter_previous': previous_id,
+                'filter_next': next_id,
+                'filter_previous_query': f'?next={next_query}&index={index - 1}',
+                'filter_next_query': f'?next={next_query}&index={index + 1}',
+            })
```
hakhalid11, line 351: I think you need to add something similar to the post method on line 366. The output returned by the post function does not have filter_count which is causing django to not render your navigation block

cantsin: Thanks! I abstracted away this functionality so I could apply it for the error edge cases. Should work better now (and the PR is cleaner as a result :tada: )

PR #667, opened 2020-09-08
create Profile model for 621

hakhalid11:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/695)

## What does this change?
Creates a Profile Model with a table migration as well as a new admin profile update create functionality
### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/models.py
```diff
@@ -32,6 +32,11 @@
 User = get_user_model()
 
 
+class Profile(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    intake_filters = models.TextField(max_length=500, blank=True)
+
```
Jkrzy, line 38: @hakhalid11, good opportunity here to provide a `__str__` method so there's a useful string representation of these instances.

Jkrzy (changes requested):
Let's give this new model class a `__str__` method, but looks good otherwise!

crt_portal/cts_forms/models.py
```diff
@@ -36,6 +36,9 @@ class Profile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE)
     intake_filters = models.TextField(max_length=500, blank=True)
 
+    def __str__(self):
+        return self.intake_filters
```
Jkrzy, line 40: This renders the contents of the `intake_filters` attribute (example below), which isn't particularly helpful since it will be modified and vary over time. I'd suggest `user` so that a developer or admin can see to which user each profile is associated.

Screen Shot 2020-09-08 at 4 13 15 PM

hakhalid11: Would you recommend the something like self.user or self.user._str_

hakhalid11:
@Jkrzy fixed `TypeError`in `_str_`

Jkrzy (approved):

PR #671, opened 2020-09-10
621 post view

hakhalid11:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/697)

## What does this change?
Creates a way to handle form submission for profile form in views.py
### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/views.py
+++ b/crt_portal/cts_forms/views.py
@@ -17,7 +17,7 @@ from formtools.wizard.views import SessionWizardView
 
 from .filters import report_filter
 from .forms import (CommentActions, ComplaintActions, ResponseActions,
-                    PrintActions, ContactEditForm, Filters,
+                    PrintActions, ContactEditForm, Filters, ProfileForm,
                     ReportEditForm, Review, add_activity)
 from .model_variables import (COMMERCIAL_OR_PUBLIC_PLACE_DICT,
                               CORRECTIONAL_FACILITY_LOCATION_DICT,
@@ -33,7 +33,7 @@ from .model_variables import (COMMERCIAL_OR_PUBLIC_PLACE_DICT,
                               LANDING_COMPLAINT_CHOICES_TO_HELPTEXT,
                               PUBLIC_OR_PRIVATE_EMPLOYER_DICT,
                               PUBLIC_OR_PRIVATE_SCHOOL_DICT)
-from .models import CommentAndSummary, Report, Trends
+from .models import CommentAndSummary, Report, Trends, Profile
 from .page_through import pagination
 
 SORT_DESC_CHAR = '-'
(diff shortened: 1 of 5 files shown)


hakhalid11:
@Jkrzy I finally got over the test error. The PR is ready for review thanks for the help

crt_portal/cts_forms/views.py
```diff
@@ -344,19 +342,15 @@ class ProfileView(LoginRequiredMixin, FormView):
 
     def post(self, request):
         """Update or create Profile"""
-        user = request.user
-        intake_filter = request.POST.get('intake_filters')
-
-        if intake_filter:
-            instance = get_object_or_404(Profile, id=user.id)
+        if request.user.profile:
```
Jkrzy, line 345: We'll need to handle this situation a bit more carefully. As written, this would never evaluate to `False`. It would throw an `RelatedObjectDoesNotExist` exception if no profile existed.

crt_portal/cts_forms/views.py
```diff
@@ -336,6 +336,103 @@ def serialize_data(report, request, report_id):
     return output
 
 
+class ProfileView(LoginRequiredMixin, FormView):
+    """Can be used for updating section filter for a profile"""
+    form_class = ProfileForm
+
+    def post(self, request):
+        """Update or create Profile"""
+        if request.user.profile:
```
Jkrzy, line 345: We'll need a more robust check here. As written, this will never evaluate to `False`. 

If a User has no associated profile, this will throw an `RelatedObjectDoesNotExist` exception and return a 500 response.

crt_portal/cts_forms/views.py
```diff
@@ -336,6 +336,103 @@ def serialize_data(report, request, report_id):
...
+                if len(p_class_list) > 3:
+                    p_class_list = p_class_list[:3]
+                    p_class_list[2] = f'{p_class_list[2]}...'
+
+                data.append({
+                    "report": report,
+                    "report_protected_classes": p_class_list,
+                    "url": f'{report.id}?next={all_args_encoded}&index={paginated_offset + index}',
+                })
+
```
Jkrzy, line 421: This looks to be a copy/paste from the current `IndexView` implementation. Where we need functionality that exists elsewhere in the code we should decompose and pull into a function that can be called in multiple places. Please do that here.

Jkrzy (changes requested):

crt_portal/cts_forms/views.py
```diff
@@ -336,6 +337,38 @@ def serialize_data(report, request, report_id):
     return output
 
 
+class ProfileView(LoginRequiredMixin, FormView):
+    """Can be used for updating section filter for a profile"""
+    form_class = ProfileForm
+
+    def has_related_profile(self, user):
+        return hasattr(user, 'profile') and user.profile is not None
```
Jkrzy, line 345: I don't think this is being executed anywhere, can be removed if so.

crt_portal/cts_forms/views.py
```diff
@@ -336,6 +337,38 @@ def serialize_data(report, request, report_id):
...
+
+    def post(self, request):
+        """Update or create Profile"""
+        if hasattr(request.user, 'profile') and request.user.profile is not None:
+            instance = request.user.profile
+        else:
+            instance = Profile()
+            instance.user = request.user
```
Jkrzy, line 353: This works but all of the evaluation is handled by `hasattr`, Opportunity here to shorten this with something like
```suggestion
        if hasattr(request.user, 'profile'):
            instance = request.user.profile
        else:
            instance = Profile(user=request.user)
```

Jkrzy (approved):

PR #676, opened 2020-09-17
Multi-select reports

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/575)

## What does this change?

Allows user to multi-select reports and bulk-assign them.

**Notes**
- The records table being squished is a known issue; we are removing the routed column shortly, which should help.
- I suggest reviewing with [the Github whitespace filter on](https://github.com/usdoj-crt/crt-portal/pull/676/files?diff=unified&w=1) as the complaints table was wrapped in a form
- The accessible autocomplete library is buggy; I have worked around this as much as possible. One known issue that I can't work around: removing the current selection still keeps the previous selection alive (until a new selection is made).

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_assign.html
+++ b/crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_assign.html
@@ -19,25 +19,43 @@
       <input type="hidden" value="{{ ids }}" name="ids" id="ids" />
       <input type="hidden" value="{{ selected_all }}" name="all" id="all" />
 
-      {{ assign_form }}
+      <div id="assign_section">
+        {{ assign_form }}
 
-      <button class="usa-button"
-              disabled
-              type="submit"
-              name="selected_ids"
-      >
-        Apply changes to {{ ids_count }} records
-      </button>
+        <button class="usa-button"
+                disabled
+                type="submit"
+                name="selected_ids"
+        >
+          Apply changes to {{ ids_count }} records
+        </button>
+
+        {% if selected_all %}
+          <button class="usa-button button--warning display-flex align-center"
+                  id="show_warning_section"
+                  disabled
+          >
+            Apply changes to {{ all_ids_count }} records
+          </button>
+        {% endif %}
+      </div>
 
       {% if selected_all %}
-        <button class="usa-button button--warning display-flex align-center"
-                type="submit"
-                disabled
-                name="all"
-                value="all"
-        >
-          Apply changes to {{ all_ids_count }} records
-        </button>
+        <div id="warning_section" hidden>
+          <p>
+            Are you sure you want to assign {{ all_ids_count }} records to <span id="warning_section_assignee"></span>?
+          </p>
+          <button class="usa-button button--warning display-flex align-center"
+                  name="all"
+                  value="all"
+                  type="submit">
+            Yes
+          </button>
+          <button class="usa-button"
+                  id="cancel_warning_section">
+            No
+          </button>
+        </div>
       {% endif %}
     </form>
   {% else %}
--- a/crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_assign.html
+++ b/crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_assign.html
@@ -11,27 +11,31 @@
 {% endblock %}
 
 {% block card_content %}
-  <form class="usa-form" method="post" novalidate>
-    {% csrf_token %}
+  {% if ids_count %}
+    <form class="usa-form" method="post" novalidate>
+      {% csrf_token %}
 
-    {{ assign_form }}
+      {{ assign_form }}
 
-    <button class="usa-button"
-            disabled
-            type="submit"
-            name="selected_ids"
-    >
-      Apply changes to {{ ids | length }} records
-    </button>
-
-    {% if selected_all %}
-      <button class="usa-button button--warning display-flex align-center"
-              type="submit"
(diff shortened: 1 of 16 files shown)


crt_portal/cts_forms/templates/forms/complaint_view/index/complaints_table.html
```diff
@@ -1,136 +1,180 @@
...
+        <tr>
+          <th>
+            <div class="usa-checkbox td-checkbox">
+              <input type="checkbox"
+                     name="all"
+                     class="usa-checkbox__input"
+                     value="all"
+                     id="checkbox-all"
+              >
+              <label class="crt-checkbox__label" for="checkbox-all"></label>
```
Jkrzy, line 22: Ooo, this generates an a11y scanning failure I think due to the label itself it empty.

crt_portal/cts_forms/templates/forms/complaint_view/index/complaints_table.html
```diff
@@ -1,136 +1,180 @@
...
+            <tr class="tr-status-{{ datum.report.status }} tr--hover{% cycle ' stripe' '' %}">
+              <td>
+                <div class="usa-checkbox td-checkbox">
+                  <input type="checkbox"
+                         name="id"
+                         class="usa-checkbox__input"
+                         value="{{ datum.report.id }}"
+                         id="checkbox-{{ datum.report.id }}"
+                  >
+                  <label class="crt-checkbox__label" for="checkbox-{{ datum.report.id }}"></label>
```
Jkrzy, line 48: Ooo, this generates an a11y scanning failure I think due to the label itself it being empty.

crt_portal/cts_forms/views.py
```diff
@@ -448,6 +452,101 @@ def post(self, request, id):
...
+
+    def post(self, request):
+        assign_form = BulkAssign(request.POST)
+        return_url_args = request.POST.get('next', '')
+        selected_all = request.POST.get('all', '') == 'all'
+        ids = request.POST.get('ids', '').split(',')
+
+        if assign_form.is_valid():
+            assignee = assign_form.cleaned_data['assigned_to']
+            requested_query = None
```
Jkrzy, line 502: This doesn't look necessary since we're assigning a value in any situation immediately thereafter.

crt_portal/static/js/complaint_actions.js
```diff
@@ -0,0 +1,40 @@
...
+  var select_all_checkboxes = dom.getElementById('checkbox-all');
+  select_all_checkboxes.onclick = function(event) {
+    var checked = event.target.checked;
+    for (var i = 0; i < all_checkboxes.length; i++) {
+      var checkbox = all_checkboxes[i];
+      if (checkbox.checked !== checked) {
+        checkbox.click(); // trigger onclick function
+      }
+    }
+  };
```
Jkrzy, line 39: If a user de-checks a checkbox after hitting `select all` -- do we need to un-check `select all`?

crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_assign.html
```diff
@@ -0,0 +1,64 @@
...
+{% endblock %}
+
+{% block card_content %}
+  {% if ids_count %}
+    <form class="usa-form" method="post" novalidate>
+      {% csrf_token %}
+
+      <input type="hidden" value="{{ return_url_args }}" name="next" id="next" />
+      <input type="hidden" value="{{ ids }}" name="ids" id="ids" />
+      <input type="hidden" value="{{ selected_all }}" name="all" id="all" />
```
Jkrzy, line 20: Do we need this hidden input for `all` if the intention is to pass it thru to the `POST` with the "apply to all reports button"?

As-is, this results in applying changes to **all** reports when hitting **either** submission button on the form.

Jkrzy (changes requested):
:amaze: @cantsin! Left a handful of comments in-line.

There's one piece where it looks like the expected functionality is off and I think we'll need some tweaks. Happy to pair and show you what I'm seeing if helpful!

crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_assign.html
cantsin: We need to preserve the `selected_all` in case of form error. That way, if the POST should fail, the user will still have the option to "select all". 

And, you are right. I was re-using the `all` name when I didn't have to; the "confirm yes" button now has the `confirm_all` name attribute. Good catch! I'll also add tests for this case ("select all but only confirm 15").

crt_portal/static/js/complaint_actions.js
cantsin: Yep! Adding this, thanks!

crt_portal/cts_forms/views.py
cantsin: Thanks! This was a leftover from when it was a large if then statement. Fixed.

crt_portal/cts_forms/templates/forms/complaint_view/index/complaints_table.html
cantsin: Thanks! :man_facepalming:

Jkrzy (approved):
Thanks for the quick turn around @cantsin! 

🚀

PR #685, opened 2020-09-24
Filter by reported issue

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/623)

## What does this change?

Adds six new filters for proforms (hidden under a `More`/`Less` button toggle):
- Primary Issue
- Reported Reason
- Relevant Details
- Hate Crime (_note, this was already implemented in the backend_)
- Service Member (_note, this was already implemented in the backend_)
- Intake Type

### Author

### Reviewer

## Notes for reviewer:


crt_portal/static/sass/custom/intake.scss
```diff
@@ -158,9 +158,21 @@
...
+      width: 20rem;
+    }
+
+    .reported-reason {
+      width: 18rem;
+    }
+
+    .relevant-details {
+      width: 18rem;
+    }
```
cantsin, line 171: Note that we have known display issues with these embiggened dropdows on IE11. Noting this for the record but it shouldn't be a problem, assuming proforms staff are indeed on Edge.

Jkrzy:
Diving into this now, initial reaction was surprise to the buttons rendering in the middle of the inputs on larger widths-- is that the desired layout? 
Screen Shot 2020-09-28 at 3 12 27 PM

cantsin:
> is that the desired layout?

Ahh yes. I think I should force a break, but let me quickly ping Jacklynn when we have our dev and design meeting later today. Note that the original issue specifies three lines with the "more" button on the _first_ line -- even though the third line is the only section that is toggled on/off though.

crt_portal/cts_forms/forms.py
```diff
@@ -951,6 +965,49 @@ class Filters(ModelForm):
...
     )
+    primary_complaint = MultipleChoiceField(
+        required=False,
+        choices=PRIMARY_COMPLAINT_PROFORM_CHOICES,
+        widget=UsaCheckboxSelectMultiple(attrs={
+            'name': 'primary_issue',
+        }),
+        label='Primary Issue',
+    )
+    reported_reason = MultipleChoiceField(
```
Jkrzy, line 976: Could be this be a `ModelMultipleChoiceField` like we're using for the `ProtectedClassForm`?

As-is this form generates unexpected escaping with:
1. Add a `reported_reason` search term
    * `race/color` results in URL param`reported_reason=race%2Fcolor` 
2. Modify another search term to search again
    * `reported_reason=race%2Fcolor` now becomes `reported_reason=race%252Fcolor`

Jkrzy (changes requested):
Thanks @cantsin! Left one comment in line that would be good to address re: unexpected escaping of search terms

cantsin:
Update re: design: `Jacklynn: Keep the first row stable, and overflow other rows so the actions (apply/more) stay in the same place`. I'll tweak accordingly (first row will always be stable with the apply and more buttons at the very right, but the rest of the filters, including the "more" filters, can reflow as needed).

cantsin:
Actually, due to the way we construct the CSS it's not easy to do the above; I think as a compromise we can lock in the first and second rows, and let the third row reflow untrammeled.

crt_portal/cts_forms/forms.py
cantsin: Huh, this is an interesting one. I don't know if we can use a `ModelMultipleChoiceField` here, it seems to make things worse. 



I don't think it has anything necessarily to do with the form or widget, I think there's something else happening here. Investigating...

Jacklynn: If I'm remembering correctly I think we're using a shorthand of these for the details page, maybe we could use that data here instead of the public phrase?

cantsin: Yep, the bug here turned out to be in the javascript code itself: we were not URI decoding query params that came in (but we were certainly URI encoding params that went out). I think this is fixed, so resolving.

Jkrzy (approved):

PR #694, opened 2020-09-29
Bulk actions

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/723)

## What does this change?

Allows the user to bulk edit:

- assigned section
- status
- primary statute
- district
- assignee (no change, this functionality was here before)
- summary

Note that a comment is required for bulk edit. 

### Author

### Reviewer

## Notes for reviewer:


cantsin:
@Jkrzy @hakhalid11 opening this up for review lest this PR become prohibitively large. Please don't forget to test on Edge. This feature is sufficiently complicated enough that I'm not sure I've tested all permutations and possible edge cases. 

Some functionality that is _missing_:

- Prefilled form fields (I believe this is best done as a separate PR)
- Tweaks such as, if assigned section is changed by itself, reset the assignee and section (again, separate PR)

crt_portal/cts_forms/views.py
```diff
@@ -524,46 +525,70 @@ def get(self, request):
...
+            if summary_string:
+                kwargs = {
+                    'is_summary': True,
+                    'note': summary_string,
+                    'author': request.user.username,
+                }
+                comment = CommentAndSummary.objects.create(**kwargs)
+                for report in requested_query:
+                    report.internal_comments.add(comment)
+                    add_activity(request.user, 'Added summary: ', summary_string, report)
```
cantsin, line 574: I debated on whether or not to override the original summary if extant but I think here it's best added as a entirely new summary, so that the last summary is still available in the database if needed. Note that `Report.get_summary` gets the last modified summary anyway so things should just continue to work. @Jkrzy curious as to your thoughts.

cantsin: I think we may have to revise this -- _editing_ a summary for a bulk updated report will change the summary for all the other (bulk updated) reports.

Two options:
- Create a new summary for each report instead (will not overwrite previous summary)
- Get or create summary for each report (this will overwrite previous summary)

Jkrzy: Good catch -- each `Report` needs it's own `Summary` instance and this should be modeled as a `ForeignKeyField` instead of `ManyToManyField`.

Option 2 you proposed is how we handle updating individual `Reports`.
We should use the same process here,  found in `ReportEditForm.save` using `CommentAndSummary.objects.update_or_create`

Jkrzy (changes requested):
Looks :amaze: overall @cantsin!

Need to address the summary creation as you noted before merging.
There's a-lot happening in `ActionsView.post` and an opportunity to shift some of that work to the `BulkActions` form

crt_portal/cts_forms/views.py
cantsin: Sure thing! Thanks -- I copied the process there for consistency.

cantsin:
> There's a-lot happening in ActionsView.post and an opportunity to shift some of that work to the BulkActions form

Agree. Moved the update functionality to the form, I think this makes the `post` functionality clearer, but do let me know what you think.

crt_portal/cts_forms/forms.py
```diff
@@ -1262,12 +1262,155 @@ class PrintActions(Form):
...
+        if comment_string:
+            kwargs = {
+                'is_summary': False,
+                'note': comment_string,
+                'author': user.username,
+            }
+            comment = CommentAndSummary.objects.create(**kwargs)
+            for report in reports:
+                report.internal_comments.add(comment)
+                add_activity(user, 'Added comment: ', comment_string, report)
```
Jkrzy, line 1394: This jumps out since we're creating a single `CommentAndSummary` instance and then associating it with _all_ of the modified reports.  Comment's are not editable so we won't encounter the bug you hit for Summaries but this feels like we may run into problems/confusion in the future.

A broader discussion, should we move towards separating `CommentAndSummary` into two separate models which are ForeignKeys to `Report`?

cantsin: Right. If we don't use this `internal_comments` field as intended (that is, as m2m) then what's the point?

I'm happy to create one for each report here if that makes things simpler down the road. Untangling the m2m relationship here into two FKs may not be too difficult, but admittedly I haven't looked at this in depth yet. Let's make a separate ticket? Meanwhile I'll update this PR to do one comment per report.

cantsin: (Updated to create one comment per report)

crt_portal/cts_forms/views.py
```diff
@@ -530,46 +531,38 @@ def get(self, request):
...
-                description = f'Updated from "{original}" to "{assignee}"'
-                add_activity(request.user, "Assigned to:", description, report)
-
-            number = requested_query.update(assigned_to=assignee)
-
-            description = f"{number} records have been assigned to {assignee}"
-            messages.add_message(request, messages.SUCCESS, description)
+            number = bulk_actions_form.update(requested_query, request.user)
+            description = bulk_actions_form.get_update_description()
+            message = f'{number} records have been updated: {description}'
```
Jkrzy, line 553: Small but handling the singular `record` case would be nice here 😄

crt_portal/cts_forms/forms.py
```diff
@@ -1262,12 +1262,155 @@ class PrintActions(Form):
...
+            for report in reports:
+                summary = report.get_summary
+                if summary:
+                    CommentAndSummary.objects.update_or_create(id=summary.id, defaults=kwargs)
+                else:
+                    summary = CommentAndSummary.objects.create(**kwargs)
+                    report.internal_comments.add(summary)
+                add_activity(user, 'Added summary: ', summary_string, report)
+
+        updated_number = reports.update(**updated_data)
```
Jkrzy, line 1412: Using `update` is OK here but since we're not calling `.save` we'll need to modify `Report.modified_date` ourselves. As-is bulk updates do not update the modified date for changed `Reports`

crt_portal/cts_forms/forms.py
Jkrzy: Agreed, we'll add a new card to:

Consider and make a decision on breaking down `CommendAndSummary`

crt_portal/cts_forms/views.py
cantsin: Sure thing, also updated the template (with the exception of "select all"). Does make me wonder if staff will use this function for one record instead of clicking on the full record.

crt_portal/cts_forms/forms.py
cantsin: Great catch! Thanks.

Jkrzy (approved):
Thanks for the quick revisions @cantsin! 

Overall this looks great! There's some opportunities to optimize the processing of these requests, but let's tackle that separately once the functionality is in. As-is, a large (10000+ record) bulk-action will likely take longer than we'd like or a user would expect to process.

PR #698, opened 2020-10-07
[WIP] new 404 error page

hakhalid11:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/705)

## What does this change?
This changes application 404 error page to include new content items such as links and text.

### Author

+ [X] Check for [accessibility](/docs/a11y_plan.md).
+ [X] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/templates/forms/error_404.html
+++ b/crt_portal/cts_forms/templates/forms/error_404.html
@@ -1,0 +1,43 @@
+{% extends "base.html" %}
+{% load i18n %}
+{% load static %}
+
+{% block page_header %}
+<header class="confirmation-header">
+  <div class="grid-container">
+    <div class="grid-row grid-gap flex-child">
+      <div class="tablet:grid-col-9 tablet:grid-offset-1">
+        <div class="doj-brand display-flex flex-align-center font-serif-lg">
+          <img src="{% static "img/doj-logo-footer.svg" %}" alt="" height="38" class="margin-right-105" />
+          <span>United States Department of Justice<span>
+        </div>
+        <h1 class="padding-top-2 padding-bottom-1 margin-0 title">
+          Civil Rights Division
+        </h1>
+      </div>
+    </div>
+  </div>
+</header>
+{% endblock %}
+
+{% block content %}
+<section class="grid-container">
+  <div class="grid-row">
+  <div class="tablet:grid-col-8 tablet:grid-offset-2">
+    <div class="crt-success-card">
+      <div class="crt-success-card__content">
+          <h2>404 | Page not found <img src="{% static 'img/alerts/error-red.svg' %}" alt="Error indicator with exclamation mark in red circle" width="50px" height="50px" class="padding-top-3"></h2>
+          <h3>We can't find the page you are looking for</h3>  
+          <p>Try one of these links:</p>
+          <div class="padding-left-1"><a href="../report/">File a complaint</a></div>
+          <div class="padding-left-1"><a href="mailto:Ask.CRT@crt.usdoj.gov">email</a>, <a href="../#contact-crt">phone or mail</a></div>
+          <div class="padding-left-1"><a href="../#about-the-division">About the Civil Rights Division</a></div>
+          <p>if you or someone else is in immediate danger, please call 911 or your local police.</p>
+      </div>
+      </div>
+    </div>
+  </div>
+  </div>
+</section>
+{% endblock %}
+
--- a/crt_portal/cts_forms/views.py
+++ b/crt_portal/cts_forms/views.py
@@ -70,11 +70,7 @@ def error_403(request, exception=None):
 def error_404(request, exception=None):
     return render(
         request,
-        'forms/errors.html', {
-            'status': 404,
-            'message': _("We can't find the page you are looking for"),
-            'helptext': _("Try returning to the previous page")
-        },
+        'forms/error_404.html',
         status=404
     )
 
--- a/crt_portal/cts_forms/tests/tests.py
+++ b/crt_portal/cts_forms/tests/tests.py
@@ -179,7 +179,7 @@ class Complaint_Show_View_404(TestCase):
         response = self.client.get(reverse('crt_forms:crt-forms-show', kwargs={'id': '1'}))
         self.assertEqual(response.status_code, 404)
         # test for custom message
-        self.assertTrue("We can&#39;t find the page you are looking for" in str(response.content))
+        self.assertTrue("404 | Page not found" in str(response.content))
 
 
 class Complaint_Show_View_Valid(TestCase):


crt_portal/cts_forms/templates/forms/error_404.html
```diff
@@ -0,0 +1,41 @@
...
+  <div class="grid-row">
+  <div class="tablet:grid-col-8 tablet:grid-offset-2">
+    <div class="crt-success-card">
+      <div class="crt-success-card__content">
+          <h1>404 | Page not found <img src="{% static 'img/alerts/error-red.svg' %}" alt="Error indicator with exclamation mark in red circle" width="50px" height="50px" class="padding-top-3"></h1>
+          <h3>We can't find the page you are looking for</h3>  
+          <p>Try one of these links:</p>
+          <div class="padding-left-2"><a href="https://civilrights.justice.gov/report/">File a complaint</a></div>
+          <div class="padding-left-2"><a href="mailto:Ask.CRT@crt.usdoj.gov">Contact us by email</a>, <a href="https://civilrights.justice.gov/#contact-crt">phone or mail</a></div>
+          <div class="padding-left-2"><a href="https://civilrights.justice.gov/ #about-the-division">About the Civil Rights Division</a></div>
```
cantsin, line 32: Please remove the space here: `https://civilrights.justice.gov/ #about-the-division`

crt_portal/cts_forms/templates/forms/error_404.html
```diff
@@ -0,0 +1,41 @@
...
+          <p>If you or someone else is in immediate danger, please call 911 or your local police.</p>
+      </div>
+      </div>
```
cantsin, line 35: Can you please re-indent this section? (I know the original error page wasn't indented ... :sweat_smile: )

crt_portal/cts_forms/views.py
```diff
@@ -70,11 +70,7 @@ def error_403(request, exception=None):
 def error_404(request, exception=None):
     return render(
         request,
-        'forms/errors.html', {
-            'status': 404,
-            'message': _("We can't find the page you are looking for"),
-            'helptext': _("Try returning to the previous page")
-        },
+        'forms/error_404.html',
```
cantsin, line 73: Wondering why we chose to make a new error page instead of re-using what we have (`forms/errors.html`)?

It looks like we have spanish translations (and the wording is the same) so we could wrap these sentences into `trans` blocks.

cantsin (changes requested):
Generally looks good, I have some minor suggestions and one big question.

crt_portal/cts_forms/views.py
hakhalid11: @cantsin I thought about this as well. One of reasons I decided on a separate file is in the acceptance criteria nice to have some of those items would be better implemented in a separate page. Plus looking at cards 706 and 707 on ZenHub, these cards have different design mocks. However, Looking at them closer I can see that we can still just update our current errors.html template and use it for this card and 706 and 707. I'd love to hear your feedback on if this is the better implementation going forward.

cantsin: I like the idea of updating our `errors.html` template because that will improve all errors in one fell swoop, and not just the 404 error. If we could do that, that'd be great! (Also I think doing so may make 706/707 work easier?)

An alternative is to add a custom block in `errors.html` that we can override, if the template doesn't work for us for whatever reason. Then we can do `{% extends error.html %}` in `error_404.html` and add the new block.

hakhalid11: Thanks for the suggestion. I spoke with Randy and Bill about what the general look and feel should be and they suggested that card 706, and 707 mocks is what the goal is. So with that in mind I updated the styling as well as moving the 404 back to using `errors.html`.

crt_portal/cts_forms/templates/forms/errors.html
```diff
@@ -3,17 +3,17 @@
...
 {% block page_header %}
-<header class="confirmation-header">
+<header class="confirmation-header crt-landing--blue padding-bottom-10">
   <div class="grid-container">
     <div class="grid-row grid-gap flex-child">
       <div class="tablet:grid-col-9 tablet:grid-offset-1">
         <div class="doj-brand display-flex flex-align-center font-serif-lg">
-          <img src="{% static "img/doj-logo-footer.svg" %}" alt="" height="38" class="margin-right-105" />
-          <span>United States Department of Justice<span>
+          <img src="{% static "img/doj-logo-footer.svg" %}" alt="" height="55" class="margin-right-105" />
```
cantsin, line 11: We do need to fill out the alt tag here. Surprised our a11y tests don't flag this. (to be clear, this isn't your fault, but since we're updating this part, let's toss in a description while we're at it)

crt_portal/static/sass/styles.scss
```diff
@@ -50,3 +50,4 @@
 @import 'custom/links';
 @import 'custom/modal';
 @import 'custom/print';
+@import 'custom/error.scss';
```
cantsin, line 53: No need for the `.scss` extension

crt_portal/cts_forms/templates/forms/errors.html
```diff
@@ -23,17 +23,46 @@ <h1 class="padding-top-2 padding-bottom-1 margin-0 title">
...
+              <li>U.S. Department of Justice</li>
+              <li>Civil Rights Division</li>
+              <li>950 Pennsylvannia Avenue, NW</li>
+              <li>Washington, D.C. 20530-0001</li>
+            </span>
+          </div>
+          <h2 class="error-subheading">Other information</h2>
+          <p><a href="https://civilrights.justice.gov/report">File a complaint</a><p>
+          <p><a href="https://civilrights.justice.gov/#about-the-division">About the Civil Rights Division</a><p>
+          <p>If you or someone else is in immediate danger, please call 911 or your local police.</p>
```
cantsin, line 61: Might want to add a tel:911 link here? We also have this in the landing page, so we can copy the translation verbatim as well:

```
{% trans 'If you or someone else is in immediate danger, <a href="tel:911">please call 911 or local police.</a>' %}
```

cantsin (changes requested):
Looking good! I think there's some scope to add the translations where we can, because some of the wording seems exactly the same. Some minor tweaks, almost there!

hakhalid11:
Thanks @cantsin, I added a some more translations that I found applied to my error page.

crt_portal/static/sass/custom/error.scss
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+.error-subheading {
+  margin: 0;
+  font-weight: $theme-font-weight-bold;
+  color: color($theme-color-primary-darker);
+}
+
+.crt-error-card {
```
cantsin, line 21: Oh, I missed this but if we are copying from another source, we can do `@extend .crt-portal-card` in SCSS.

cantsin (changes requested):
Overlooked one possible thing in the SCSS. Rest looks good! If you didn't copy from another source for the SCSS, then let's go ahead and merge.

cantsin (approved):
LGTM!

PR #699, opened 2020-10-07
Bulk actions prefill

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/728)

## What does this change?

- Pre-fills bulk action form iff the queryset has same values.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -756,12 +756,12 @@ class ProForm(
         Contact.__init__(self, *args, **kwargs)
         # CRT views only
         self.fields['intake_format'] = ChoiceField(
-            choices=_add_empty_choice((
+            choices=(
                 ('letter', 'letter'),
                 ('phone', 'phone'),
                 ('fax', 'fax'),
                 ('email', 'email'),
-            )),
+            ),
             widget=UsaRadioSelect,
             required=False,
         )
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1262,7 +1262,7 @@ class PrintActions(Form):
     )
 
 
-class BulkActions(Form, ActivityStreamUpdater):
+class BulkActionsForm(Form, ActivityStreamUpdater):
     assigned_section = ChoiceField(
         label='Section',
         widget=ComplaintSelect(
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1263,7 +1263,7 @@ class PrintActions(Form):
 
 
 class BulkActionsForm(Form, ActivityStreamUpdater):
-    EMPTY_CHOICE ='Multiple'
+    EMPTY_CHOICE = 'Multiple'
     assigned_section = ChoiceField(
         label='Section',
         widget=ComplaintSelect(
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1158,7 +1158,7 @@ class ComplaintActions(ModelForm, ActivityStreamUpdater):
                     'class': 'text-uppercase crt-dropdown__data',
                 },
             ),
-            choices=_add_empty_choice(STATUTE_CHOICES),
+            choices=_add_empty_choice(STATUTE_CHOICES, default_string=''),
             required=False
         )
         self.fields['district'] = ChoiceField(
@@ -1168,7 +1168,7 @@ class ComplaintActions(ModelForm, ActivityStreamUpdater):
                     'class': 'text-uppercase crt-dropdown__data',
                 },
             ),
-            choices=_add_empty_choice(DISTRICT_CHOICES),
+            choices=_add_empty_choice(DISTRICT_CHOICES, default_string=''),
             required=False
         )
         self.fields['assigned_to'].widget.label = 'Assigned to'
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1329,9 +1329,12 @@ class BulkActionsForm(Form, ActivityStreamUpdater):
         singular value within that query. Used to set initial fields
         for bulk update forms.
         """
+        # make sure the queryset does not order by anything, otherwise
+        # we will have difficulty getting distinct results.
+        query = record_query.order_by()
         for key in keys:
-            values = record_query.values_list(key, flat=True)
-            if len(set(values)) == 1:
+            values = query.order_by().values_list(key, flat=True).distinct()
+            if values.count() == 1:
                 yield key, values[0]
 
     def __init__(self, query, *args, **kwargs):
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1333,7 +1333,7 @@ class BulkActionsForm(Form, ActivityStreamUpdater):
         # we will have difficulty getting distinct results.
         query = record_query.order_by()
         for key in keys:
-            values = query.order_by().values_list(key, flat=True).distinct()
+            values = query.values_list(key, flat=True).distinct()
             if values.count() == 1:
                 yield key, values[0]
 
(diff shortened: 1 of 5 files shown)


crt_portal/cts_forms/forms.py
```diff
@@ -1322,6 +1323,16 @@ class BulkActions(Form, ActivityStreamUpdater):
...
 
+    def __init__(self, query):
+        Form.__init__(self)
+        keys = ['assigned_section', 'status', 'primary_statute', 'district']
+        values = query.values_list(*keys)
+        initial_values = list(zip(*values))
+        for index, key in enumerate(keys):
+            initial = initial_values[index]
+            if len(set(initial)) == 1:
+                self.fields[key].initial = initial[0]
```
cantsin, line 1334: @Jkrzy I'm not crazy about this block of code, it's fairly complicated and it feels like we can do some fancy database magic to get these results quickly. Maybe using annotations? Open to ideas.

crt_portal/cts_forms/templates/forms/widgets/multi_select_option.html
```diff
@@ -1,3 +1,3 @@
 <option value="{{option.value}}" {% if option.selected is not False %}selected{% endif %}>
-  {{option.value|title}}
+  {{option.value|default:option.label|title}}
```
cantsin, line 2: Required because otherwise we always default to a blank value. Note that this impacts other forms in proforms as well.

crt_portal/cts_forms/forms.py
```diff
@@ -1341,7 +1351,7 @@ def get_update_description(self):
             what = value.lower()
             item = self.cleaned_data[key]
             string = custom_strings.get(what, default_string)
-            description = string.format(**{'what': what, 'item': item})
+            description = string.format(**{'what': what, 'item': item or "''"})
```
cantsin, line 1354: Whoops, we want this in case we unset a field when doing a bulk update.

crt_portal/cts_forms/forms.py
Jkrzy: This is a great candidate to be moved into it's own function with a docstring to both clarify the intent, facilitate testing and evolution to something you're more comfortable with.

Nothing immediately comes to mind from the Django ORM that would get us exactly what we're looking for -- Are all values for `X` the same across our QuerySet? -- in a single query

Breaking it into one query for each field is an option, straightforward, and avoids potential problems with the inbound queryset being huge. Something like

```python
values = Report.objects.all().values_list('status', flat=True).distinct()

print(values.query)
# SELECT DISTINCT "cts_forms_report"."status" FROM "cts_forms_report"

initial = values[0] if len(values) == 1 else None

```

crt_portal/cts_forms/views.py
```diff
@@ -536,18 +536,19 @@ def get(self, request):
...
-            if confirm_all:
-                requested_query = self.reconstruct_query(return_url_args)
-            else:
-                requested_query = Report.objects.filter(pk__in=ids)
+        if confirm_all:
+            requested_query = self.reconstruct_query(return_url_args)
+        else:
+            requested_query = Report.objects.filter(pk__in=ids)
 
+        bulk_actions_form = BulkActions(request.POST, requested_query)
```
Jkrzy: Not related to the functionality here but I'd love to rename `BulkActions` to  `BulkActionsForm` and help nudge us towards more idiomatic Django class names.

crt_portal/cts_forms/forms.py
```diff
@@ -1263,35 +1263,36 @@ class PrintActions(Form):
...
 
 class BulkActions(Form, ActivityStreamUpdater):
+
     assigned_section = ChoiceField(
         label='Section',
         widget=ComplaintSelect(
             attrs={'class': 'usa-select text-bold text-uppercase crt-dropdown__data'},
         ),
-        choices=_add_empty_choice(SECTION_CHOICES),
+        choices=_add_empty_choice(SECTION_CHOICES, default_string='Multiple'),
```
Jkrzy, line 1271: Small opportunity here to define `Multiple` in a single place and not hard code it across all of our fields here.

crt_portal/cts_forms/forms.py
```diff
@@ -1263,35 +1263,36 @@ class PrintActions(Form):
 
 
 class BulkActions(Form, ActivityStreamUpdater):
```
Jkrzy: Not related to the functionality here but I'd love to rename this for `BulkActionsForm` and help nudge the entire codebase towards more idiomatic Django class names.

Jkrzy (changes requested):
Noted a few opportunities for some technical tweaks but looks great overall!

The acceptance criteria in https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/723 don't capture this well, I'd like to demo for Bill/Randy and have them confirm this is functioning as expected and that the impacts to other forms are acceptable. 

The casing of `Multiple` differs across the inputs rendered on the bulk action page. The changes to `multi_selection_option` also impact the deport details page.

crt_portal/cts_forms/forms.py
cantsin: I agree. Renamed in this PR, but I would also suggest a follow up PR right after this one since we have other form names: `Filters`, `ResponseActions`, `ComplaintActions`, and `PrintActions` should be renamed accordingly.

Jkrzy:
I caught up with Bill/Randy just after stand-up this morning and they confirmed we should limit the impact of this implementation to the Bulk Actions form itself. 

So we'll need a bit of extra work here on the form fields so that the `empty` option on the Report Details pages are not set to `- SELECT -`

Screen Shot 2020-10-08 at 1 19 06 PM

crt_portal/cts_forms/forms.py
cantsin: Updated with a simpler implementation, thanks. `distinct` doesn't seem to work here because the queryset may already have filters applied to it previously. 

For instance:

```sql
SELECT DISTINCT "cts_forms_report"."district", "cts_forms_report"."create_date" FROM "cts_forms_report" WHERE "cts_forms_report"."status" IN (new, open) ORDER BY "cts_forms_report"."create_date" DESC
```

Gives me `[None, 2, 2, '', None, None ...]`

I think we'd have to force the evaluation of the queryset, I tried doing `all()` but that just returns a copy. I may be missing something here, though.

cantsin:
> So we'll need a bit of extra work here on the form fields so that the empty option on the Report Details pages are not set to `- SELECT -`

Done.

cantsin:
> The casing of Multiple differs across the inputs rendered on the bulk action page. 

Yes. We cannot style individual options in a `select`, unfortunately. It's a bit of all-or-nothing.

crt_portal/cts_forms/forms.py
Jkrzy: Ah, I think the `order_by` is the culprit -- https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.distinct

I don't _think_ we care about order here so this may work:
```python
query.order_by()...
```

cantsin: You rock! That was exactly it.

cantsin:
@Jkrzy I think this is ready to go now. Let me know if there's anything else that catches your eye.

crt_portal/cts_forms/views.py
```diff
@@ -517,7 +517,7 @@ def get(self, request):
         ids = request.GET.getlist('id')
         ids_count = len(ids)
 
-        bulk_actions_form = BulkActions()
+        bulk_actions_form = BulkActionsForm(requested_query)
```
Jkrzy, line 520: It looks like we need to handle this in the same way we are in `post` below. This takes the full query, regardless of selected Reports, and uses it to generate initial values. 

Resulting in `MULTIPLE` values being rendered on the form for `status` when 3 reports with an identical `status` are selected.

cantsin: Great catch! Thanks. Tossed up a fix for this.

Jkrzy (approved):
Looks good, thanks @cantsin!

PR #710, opened 2020-10-16
Customer facing Error page

hakhalid11:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/707undefined)

## What does this change?
Handles 500 error or any other subsequent error status other than error 404. Modified templating to handle 404 page and any other error message heading (Can be used for maintenance page card). Removed unused error message data being passed to template (only exception is error_404 and csrf_failure). **This page is not fully translated in Spanish specifically the card headings**.

image

### Author
hakhalid11

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/templates/forms/errors.html
+++ b/crt_portal/cts_forms/templates/forms/errors.html
@@ -29,7 +29,7 @@
           <div class="display-flex flex-column padding-bottom-3">
             <h1 class=".h1__display">{% block error_heading %}We're sorry, something went wrong{% endblock %}</h1>
             <span class="crt-error_message">{% block error_message %}We hope to fix this issue shortly{% endblock %}</span>
-            <span class="help-text__small">{{ helptext }}</span>
+            <span class="help-text__small">{% block error_helptext %}Error: {{ status }}{% endblock %}</span>
           </div>
           <h2 class="error-subheading">{% trans 'If you need to contact us immediately, you can reach us by:' %}</h2>
           <div class="crt-error-box">
--- a/crt_portal/cts_forms/templates/forms/errors.html
+++ b/crt_portal/cts_forms/templates/forms/errors.html
@@ -27,8 +27,8 @@
       <div class="crt-error-card">
         <div class="crt-error-card__content">
           <div class="display-flex flex-column padding-bottom-3">
-            <h1 class=".h1__display">{{ status }}</h1>
-            <span>{{ message }}</span>
+            <h1 class=".h1__display">{% block error_heading %}We're sorry, something went wrong{% endblock %}</h1>
+            <span class="crt-error_message">{% block error_message %}We hope to fix this issue shortly{% endblock %}</span>
             <span class="help-text__small">{{ helptext }}</span>
           </div>
           <h2 class="error-subheading">If you need to contact us immediately, you can reach us by:</h2>
--- a/crt_portal/cts_forms/templates/forms/errors.html
+++ b/crt_portal/cts_forms/templates/forms/errors.html
@@ -57,8 +57,8 @@
             </span>
           </div>
           <h2 class="error-subheading">{% trans 'Other information' %}</h2>
-          <p><a href="https://civilrights.justice.gov/report">{%trans 'File a complaint' %}</a><p>
-          <p><a href="https://civilrights.justice.gov/#about-the-division">{% trans 'About the Civil Rights Division' %}</a><p>
+          <p><a href="https://civilrights.justice.gov/report">{%trans 'File a complaint' %}</a></p>
+          <p><a href="https://civilrights.justice.gov/#about-the-division">{% trans 'About the Civil Rights Division' %}</a></p>
           <p>{% trans 'If you or someone else is in immediate danger, <a href="tel:911">please call 911 or local police.</a>' %}</p>
         </div>
       </div>
--- a/crt_portal/cts_forms/templates/forms/errors.html
+++ b/crt_portal/cts_forms/templates/forms/errors.html
@@ -29,7 +29,7 @@
           <div class="display-flex flex-column padding-bottom-3">
             <h1 class=".h1__display">{% block error_heading %}{% trans "We're sorry, something went wrong" %}{% endblock %}</h1>
             <span class="crt-error_message">{% block error_message %}{% trans "We hope to fix this issue shortly" %}{% endblock %}</span>
-            <span class="help-text__small">{% block error_helptext %}Error: {{ status }}{% endblock %}</span>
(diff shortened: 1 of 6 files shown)


crt_portal/cts_forms/views.py
```diff
@@ -82,9 +82,9 @@ def error_500(request, exception=None):
     return render(
         request,
         'forms/errors.html', {
-            'status': 500,
-            'message': _("There's a problem loading this page"),
-            'helptext': _("There's a technical problem loading this page. Try refreshing this page or going to another page. If that doesn't work, try again later.")
+            'status': "We're sorry, something went wrong",
+            'message': "We hope to fix this issue shortly",
+            'helptext': "Error: 500"
```
cantsin, line 87: From reading the original ticket, I think this covers all classes of server errors (5xx errors). Also there seems to be a spanish translation provided?

cantsin (changes requested):
I'm not sure this PR was done, because it still has `[WIP]` in the title?

crt_portal/cts_forms/views.py
hakhalid11: Yes its a WIP, I think I accidentally added reviewers.

hakhalid11:
> I'm not sure this PR was done, because it still has `[WIP]` in the title?

This PR is ready for review

crt_portal/cts_forms/views.py
```diff
@@ -48,8 +48,6 @@ def error_400(request, exception=None):
         request,
         'forms/errors.html', {
             'status': 400,
-            'message': _("Bad request"),
-            'helptext': _("It seems your browser is not responding properly. Try refreshing this page.")
```
cantsin, line 52: It strikes me that these messages and helptext are worth preserving and that we can keep them somewhere in the template. Are we deliberately moving away from specific error messages and being more generic? I'm not sure I agree with this since users will have little frame of reference other than literally, "something went wrong". How will users communicate specifics with us if an error does, in fact, arise?

crt_portal/cts_forms/templates/forms/errors.html
```diff
@@ -27,9 +27,9 @@
       <div class="crt-error-card">
         <div class="crt-error-card__content">
           <div class="display-flex flex-column padding-bottom-3">
-            <h1 class=".h1__display">{{ status }}</h1>
-            <span>{{ message }}</span>
-            <span class="help-text__small">{{ helptext }}</span>
+            <h1 class=".h1__display">{% block error_heading %}We're sorry, something went wrong{% endblock %}</h1>
+            <span class="crt-error_message">{% block error_message %}We hope to fix this issue shortly{% endblock %}</span>
+            <span class="help-text__small">{% block error_helptext %}Error: {{ status }}{% endblock %}</span>
```
cantsin, line 32: Did we change the order from `status`/`message`/`helptext`? it's a bit confusing here since the classes don't reflect this change. 

Side note, using blocks seems overly complicated for this use case -- usually we want blocks to override more complicated functionality. Here, I believe we can do something simpler: `{{ message|default:"We're sorry, something went wrong" }}` instead.

cantsin (changes requested):
I would keep the current messages and helptext (and spanish translations!) in the 5xx error pages as these are valuable information to retain imo, otherwise the user has nothing else to refer to other than "something went wrong". Happy to bounce back to Bill and Randy if needed. 

Other than that, I believe we can simplify this PR by using the [default](https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#default) template tag.

hakhalid11:
@cantsin you bring up good points, both of your change requests comments stem from the following [design mock](https://github.com/18F/crt-portal/issues/707#issuecomment-692223840). I was initially hesitant to remove the error message/helpful text statements but reading your comment I think we should get some feedback from Randy and Bill. 

Your second comment is a direct result of the new design mock, moving status to the bottom with smaller text. In general I agree with your broad point I think cleanup of template blocks can be done and I can make a sample design with helpful text and messages to show Randy and Bill. I would love any your thoughts/comments going forward

cantsin:
Thanks for being receptive, bouncing back to bill or randy with the above comments sounds good! If you tag me I can jump into the conversation and I can assist?

hakhalid11:
@cantsin Randy gave some great feedback regarding this card and he would like to stick with the original design. With that being said I looked at your example for default filter and I think we will still need a template block. Because design requires the status to be in different place depending on if its a 500 error or a 404 error I went with template blocks. I added some minor cleanup in padding and fixed a syntax typo in the html. But I think the rest of the code meets the acceptance criteria and feedback from product owner.

Jkrzy (commented):
There's some accessibility issues we missed in https://github.com/usdoj-crt/crt-portal/pull/698/ that persist here, we can tackle those in another [card](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/760).

For this PR, the translation implementation is incomplete. That's noted in the original PR but it's unclear as to why, we've been provided the translations including the card headers. @hakhalid11, @randyabramson, is there a piece of context I'm missing for _why_ the headers can't be translated right now?

hakhalid11:
@Jkrzy I missed Randy's comment regarding Spanish translations so thanks for the heads up 👍. They should be in there and good to go.

crt_portal/cts_forms/templates/forms/errors.html
```diff
@@ -27,9 +27,9 @@
       <div class="crt-error-card">
         <div class="crt-error-card__content">
           <div class="display-flex flex-column padding-bottom-3">
-            <h1 class=".h1__display">{{ status }}</h1>
-            <span>{{ message }}</span>
-            <span class="help-text__small">{{ helptext }}</span>
+            <h1 class=".h1__display">{% block error_heading %}{% trans "We're sorry, something went wrong" %}{% endblock %}</h1>
+            <span class="crt-error_message">{% block error_message %}{% trans "We hope to fix this issue shortly" %}{% endblock %}</span>
+            <span class="help-text__small">{% block error_helptext %}Error: {{ status }}{% endblock %}</span>
```
Jkrzy, line 32: Missing a translation block here for `Error`. It happens to work for ES translations but will definitely bite us in the future.

Jkrzy (changes requested):
Thanks @hakhalid11! 

One missing translation but looks good otherwise.

hakhalid11:
@Jkrzy Thanks for the heads up I added the translation for `Error:` as well

Jkrzy (approved):

PR #715, opened 2020-10-19
Bulk actions print

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/734)

## What does this change?

- Adds a new template for printing complaint details. the old complaint details page was problematic because
    1. form duplication resulted in errors in the console regarding duplicate ids and hidden named inputs
    2. a lot of data was being unnecessarily transferred over for each report
- Undoes previous print changes for complaint details
- Tries to re-use existing single print functionality where-ever possible
- Refactors refactoring query in views 
- Adds the printer svg icon (previously we were using the fax svg icon)

### Author

### Reviewer

## Notes for reviewer:


cantsin:
@Jkrzy This PR may be over the place since there's refactoring involved; I can split it up in multiple PRs if you would prefer.

crt_portal/cts_forms/views.py
```diff
@@ -310,7 +315,6 @@ def index_view(request):
 
 
 def serialize_data(report, request, report_id):
-    primary_complaint = [choice[1] for choice in PRIMARY_COMPLAINT_CHOICES if choice[0] == report.primary_complaint]
```
cantsin, line 313: I'm not sure why this was set as a array with a single item, so let me know if I'm missing anything here. (As far as I can see, we only reference this variable once, as `{{primary_complaint.0}}` in the template.)

crt_portal/cts_forms/models.py
```diff
@@ -317,6 +318,12 @@ def get_absolute_url(self):
     def closed(self):
         return self.status == CLOSED_STATUS
 
+    def primary_complaint_description(self):
```
Jkrzy, line 321: Django has a built-in method we can leverage to provide this functionality since we're specifying a `choices` attribute on the model field! 

https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.get_FOO_display

`get_primary_complaint_display` will return the human readable value of `primary_complaint`

Jkrzy (changes requested):
Looks amazing, thanks for tackling this one @cantsin!

 Noted a small few tweaks we could make before merging this into develop.

crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_print.html
```diff
@@ -0,0 +1,32 @@
+{% load static %}
+
+{% if print_reports.count <= 100 %}
```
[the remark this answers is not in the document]
cantsin: Welp! Yes. I need to go back and re-do this bit. Thanks for the catch.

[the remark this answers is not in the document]
cantsin: Updated ... :sweat_smile:

crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_print.html
```diff
@@ -0,0 +1,21 @@
...
+
+<span>
+  <a href="#" id="printout_report" class="bulk-print-link">
+    <img src="{% static "img/intake-icons/ic_printer.svg" %}" alt="printer" class="icon" />
+    Print all
+  </a>
+</span>
+
+<div class="bulk-print-reports display-none">
+  {% for report in print_reports|slice:":100" %}
```
Jkrzy, line 11: Ooo, hard coding `100` here results in us rendering 100 reports on the page _anytime_ `print_reports >= 100`. 
With the current UI, a user can only actually print those 100 reports if they've selected -- passed 100 IDs in their `GET`, or `print_reports.count() == 100`.

There's an opportunity here to save a bit of work and limit what we're rendering in the template to reports which can be printed by the end user. I'd call this an optimization and we don't need to block merging this whole PR on it but wanted to mention!

Jkrzy (approved):
Overall looks :amaze: thanks @cantsin! I left one comment inline above re: a potential area we could optimize and do a bit less rendering.

crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_print.html
cantsin: Good catch. I'm playing with this idea over at https://github.com/usdoj-crt/crt-portal/pull/739

PR #717, opened 2020-10-20
bulk actions bugfix: do not allow fields to be unset

cantsin:
[Link to Github comment.](https://github.com/18F/crt-portal/issues/723#issuecomment-713041800)

## What does this change?

Bugfix: do not allow user to select "Multiple" for either `assigned section` or `status`.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1349,6 +1349,11 @@ class BulkActionsForm(Form, ActivityStreamUpdater):
 
     def get_updates(self):
         updates = {field: self.cleaned_data[field] for field in self.changed_data}
+        # do not allow any fields to be unset. this may happen if the
+        # user selects "Multiple".
+        for key in ['assigned_section', 'status', 'primary_statute']:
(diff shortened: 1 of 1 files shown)


Jkrzy (changes requested):
I think this is reasonable for an immediate fix @cantsin.

We'll need to ignore empty values from all of the fields where `MULTIPLE` can be selected so let's include `primary_statute`.

cantsin:
> We'll need to ignore empty values from all of the fields where MULTIPLE can be selected so let's include primary_statute.

Added. 

> I think this is reasonable for an immediate fix @cantsin.

I think we'll have to do a hotfix, given that we're releasing in, oh, 1.5 hours. I would like to test this feature in either dev/staging first. What do you think?

Jkrzy:
Looking ahead, we should refine this interaction by:
* Only rendering `MULTIPLE` in the `select` element if it is the initial value
* Validate the POSTed data to preclude empty values and return a form error

Jkrzy (approved):

PR #718, opened 2020-10-21
Bulk change rules

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/758)

## What does this change?

- Prevents more than 500 bulk updates
- Changes status to "new" and unsets assignee (and disables both) if assigned section is changed

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1357,6 +1357,7 @@ class BulkActionsForm(Form, ActivityStreamUpdater):
         # if section is changed, override assignee and status
         # explicitly, even if they are set by the user.
         if 'assigned_section' in updates:
+            updates['primary_statute'] = ''
             updates['assigned_to'] = ''
             updates['status'] = 'new'
         updates.pop('district', None)  # district is currently disabled (read-only)
--- a/crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_actions.html
+++ b/crt_portal/cts_forms/templates/forms/complaint_view/actions/bulk_actions.html
@@ -28,6 +28,17 @@
           </label>
           {{ bulk_actions_form.assigned_section }}
         </div>
+        <p>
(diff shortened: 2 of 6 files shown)


Jkrzy (changes requested):
Thanks @cantsin!!!

@randyabramson and @blaughman are going to update the associated issue clarifying which fields on the Report need to be set to `None` upon change of `Section` via BulkActions.

Flagging this as `requested changes` in anticipation of those updates.

cantsin:
Side note, merging against `develop` but let me know if I should merge against `release/10-20-2020` to get into staging sooner.

randyabramson:
@cantsin I am OK with going to staging. If for some reason we are really unhappy with it, can we disable bulk actions on Tuesday so that we can still proceed with the release?

Jkrzy (changes requested):
Looks like **primary classification** is not being cleared as expected when the section is changed via bulk actions.

cantsin:
> Looks like **primary classification** is not being cleared as expected when the section is changed via bulk actions.

Ah yes, thank you. Fixed (on the backend).

Jkrzy:
I don't think this was enumerated in the acceptance criteria but we should avoid modifying `status`, `assigned_to`, and `primary_statute` if they are already the desired values. 

As is, this generates confusing entries in the activity log:

Screen Shot 2020-10-22 at 4 26 05 PM


This may be best to tackle in a new card since I think it will require moving away from using `QuerySet.update`

cc: @randyabramson, @blaughman, @cantsin

Jkrzy (approved):
Looks good! Made one comment re: a piece of functionality that I'd like to tackle in another card since this one fulfills https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/758 by itself.

PR #735, opened 2020-11-03
fix metadata syntax

LindsayYoung:
I was looking at the Lighthouse Audit for the site and I think that we need to set the metadata description with "name" rather than "property." (The og:description should be a property, so that is fine.)

https://web.dev/meta-description/?utm_source=lighthouse&utm_medium=devtools


cantsin (approved):
LGTM

PR #747, opened 2020-11-17
compress assets: re-enable, configure

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/562)

## What does this change?

Compresses JS assets. CSS is already compressed, via gulp and `nanocss`. 

From some rough benchmarks, when viewing a report (`/form/view`), the compressing of javascript itself is negligible (before: 705kb transferred; after: 692kb transferred), but the number of requests goes down from 33 to 22 and so does rendering speed, from ~4.5s to ~2s (when `DOMContentLoaded` triggers).

So overall compressing does offer some benefits. One thing to consider is that bundling (not necessarily minifying) our javascript files would help just as much.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/crt_portal/settings.py
+++ b/crt_portal/crt_portal/settings.py
@@ -335,8 +335,7 @@ COMPRESS_PRECOMPILERS = (
     ('css', 'compressor_toolkit.precompilers.SCSSCompiler'),
 )
 
-# would like to add this before public release
-COMPRESS_ENABLED = False
+COMPRESS_ENABLED = True
 
 # adding better messaging
 CSRF_FAILURE_VIEW = 'cts_forms.views.csrf_failure'
--- a/crt_portal/crt_portal/settings.py
+++ b/crt_portal/crt_portal/settings.py
@@ -250,6 +250,7 @@ if environment not in ['LOCAL', 'UNDEFINED']:
     STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_DEFAULT_ACL = 'public-read'
+    AWS_IS_GZIPPED = True
 
 if environment in ['PRODUCTION', 'STAGE', 'DEVELOP']:
     MIDDLEWARE.append('csp.middleware.CSPMiddleware')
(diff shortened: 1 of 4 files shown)


crt_portal/cts_forms/templates/forms/base.html
```diff
@@ -99,8 +100,10 @@
     </div>
     {% endblock %}
   </body>
+  {% compress js %}
   <script src="{% static 'js/url_params_polyfill.js' %}"></script>
   <script src="{% static 'js/uswds.min.js' %}"></script>
   <script src="{% static 'js/focus_alert.js' %}"></script>
   {% block page_js %}{% endblock %}
```
cantsin, line 107: Note that most of our per-template JS is in a `page_js` block, so this captures a whole lot of other linked javascript files and inline javascript.

crt_portal/crt_portal/settings.py
```diff
@@ -335,8 +335,7 @@
     ('css', 'compressor_toolkit.precompilers.SCSSCompiler'),
 )
 
-# would like to add this before public release
-COMPRESS_ENABLED = False
+COMPRESS_ENABLED = True
```
cantsin, line 338: I considered disabling this for local dev usage, but I like the increased rendering speed. Might cause some issues when debugging, but open to suggestions here.

cantsin:
Now gzipping on s3, hopefully -- but no way to test until we try in dev. (If someone knows how to test otherwise, let me know please)

Jkrzy:
Our usage of `django-compressor` feels wonky, I propose we break this one into two separate efforts/cards:

1. Bundling/Minification of assets
    * I'm not confident the `django-compressor` setup we have is going to play nice with `django-storages` and s3. 
    * Given the small benefit you noticed this feels like a much lower priority to me
2. Serving compressed static files
    * Independent of compressor, `AWS_IS_GZIPPED=True` works fine in my testing w/ `localstack`


Bonus 3: Work towards environment parity and have S3 storage be an option for local development.  https://github.com/localstack/localstack or https://www.npmjs.com/package/s3rver are options. I've got a proof of concept for localstack I can push up.

Feel reasonable @cantsin?

cantsin:
Definitely reasonable. `django-compressor` feels like its not necessary here; maybe we can remove outright, since we already use gulp to copy the js files -- maybe we can go ahead and bundle them up instead there (in `gulpfile.js`). Will do a bit of investigation.

Jkrzy:
I went ahead and cherry-picked out the gzipping into https://github.com/usdoj-crt/crt-portal/pull/750 -- @cantsin, going to close this one.

PR #751, opened 2020-11-24
Disable NR auto-instrumentation

Jkrzy:
[No card](link-goes-here)

## What does this change?
Disables auto-instrumentation by the NR python agent in our settings. We do not want NR to inject content into our
application's responses and our CSP configuration prevents it from being functional.

### Author

### Reviewer

## Notes for reviewer:


cantsin (approved):
1000% LGTM.

PR #760, opened 2020-12-04
Compress and fingerprint javascript assets

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/792)

## What does this change?

Turns on compressor (bundles javascript and css assets), and fingerprints them via ManifestFiles.

Bundling is on for all environments and provides a modest speedup when loading the page. 

I tried my best to get rid of gulp but was unable to do so, mostly because of the way USDWS is set up. I have, however, cleaned up the `gulpfile.js` and removed some stale code/libraries.

---

*Note*: Testing fingerprinting on S3 is difficult unless you already have localstack set up. Here's an alternative to test fingerprinting: put this line in local settings:

```python
COMPRESS_ENABLED = True
```

You should see something like:

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/crt_portal/settings.py
+++ b/crt_portal/crt_portal/settings.py
@@ -16,6 +16,7 @@ import os
 import boto3
 from django.utils.log import DEFAULT_LOGGING
 from django.utils.translation import gettext_lazy as _
+from django.contrib.staticfiles.storage import ManifestFilesMixin
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
@@ -229,6 +230,22 @@ if environment == 'PRODUCTION':
 STATIC_URL = '/static/'
 
 if environment not in ['LOCAL', 'UNDEFINED']:
+
+    # this import will throw errors if SECRET_KEY is not set yet in a
+    # local environment (we import local_settings at the end)
+    from storages.backends.s3boto3 import S3Boto3Storage
+
+    class ManifestS3FilesStorage(ManifestFilesMixin, S3Boto3Storage):
+        def read_manifest(self):
+            """
+            Work around a bug where S3Boto3Storage throws IOError but
+            ManifestFilesMixin expects FileNotFound.
+            """
+            try:
+                return super(ManifestS3FilesStorage, self).read_manifest()
+            except IOError:
+                return None
+
     for service in vcap['s3']:
         if service['instance_name'] == 'crt-s3':
             # Public AWS S3 bucket for the app
@@ -247,7 +264,7 @@ if environment not in ['LOCAL', 'UNDEFINED']:
     AWS_LOCATION = 'static'
     AWS_QUERYSTRING_AUTH = False
     STATIC_URL = f'https://{AWS_S3_CUSTOM_DOMAIN}/{AWS_LOCATION}/'
-    STATICFILES_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
+    STATICFILES_STORAGE = 'ManifestS3FilesStorage'
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_DEFAULT_ACL = 'public-read'
     AWS_IS_GZIPPED = True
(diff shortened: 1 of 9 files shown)


crt_portal/crt_portal/settings.py
```diff
@@ -336,9 +353,6 @@
     ('css', 'compressor_toolkit.precompilers.SCSSCompiler'),
 )
 
-# would like to add this before public release
-COMPRESS_ENABLED = False
```
cantsin, line 340: By default this is set to whatever DEBUG is, so omitting this line for convenience

gulpfile.js
```diff
@@ -81,10 +79,6 @@ gulp.task('build-sass', function(done) {
   var plugins = [
     // Autoprefix
     autoprefixer(autoprefixerOptions),
-    // Pack media queries
-    mqpacker({ sort: true }),
-    // Minify
-    cssnano(({ autoprefixer: { browsers: autoprefixerOptions }}))
```
cantsin, line 87: css-mqpacker is no longer supported -- see https://www.npmjs.com/package/css-mqpacker

cssnano is no longer needed thanks to django compressor. there _shouldn't_ be any difference between cssnano/django-compressor, and I can't see any difference in the resulting design.

crt_portal/crt_portal/settings.py
Jkrzy: 👍 , Noting ref from docs: The default value for `COMPRESS_ENABLED` is the _opposite_ of `DEBUG`

https://django-compressor.readthedocs.io/en/stable/settings/#django.conf.settings.COMPRESS_ENABLED

crt_portal/cts_forms/templates/forms/base.html
```diff
@@ -99,8 +104,10 @@
     </div>
     {% endblock %}
   </body>
+  {% compress js %}
   <script src="{% static 'js/url_params_polyfill.js' %}"></script>
   <script src="{% static 'js/uswds.min.js' %}"></script>
   <script src="{% static 'js/focus_alert.js' %}"></script>
   {% block page_js %}{% endblock %}
```
Jkrzy, line 111: We've got some external sources in templates that use this `page_js` block that end up leading to an exception when testing locally -- touchpoints script in `report_primary_complaint.html` and `confirmation.html` specifically

```
UncompressableFileError at /report/
'https://touchpoints.app.cloud.gov/touchpoints/e0517981/js' isn't accessible via COMPRESS_URL ('/static/') and can't be compressed
```

Jkrzy (changes requested):
Thanks @cantsin! In testing locally this breaks report submission, left one comment in-line where I think the issue lies.

crt_portal/cts_forms/templates/forms/base.html
cantsin: Gosh, great catch, thank you!!

cantsin: Fixed + rebased on dev

crt_portal/cts_forms/templates/forms/confirmation.html
```diff
@@ -189,7 +189,10 @@ <h2 class='h1__display'>{% trans "Your submission" %}</h2>
 {% endblock footer_extra %}
 
 {% block page_js %}
-  <script src="https://touchpoints.app.cloud.gov/touchpoints/a24aeb5e/js" async></script>
   <script src="{% static 'js/modal.js' %}"></script>
   <script src="{% static 'js/redirect-modal.js' %}"></script>
 {% endblock %}
+
+{% block touchpoint_js %}
```
Jkrzy, line 196: Looks like the implementation in `report_primary_complaint.html` here differs by one character so it doesn't extend this block. 
```suggestion
{% block touchpoints_js %}
```

crt_portal/cts_forms/templates/forms/base.html
```diff
@@ -99,8 +104,13 @@
...
+  {% endcompress %}
+
+  {# do not compress #}
+  {% block touchpoint_js %}{% endblock %}
```
Jkrzy, line 115:
```suggestion
  {% block touchpoints_js %}{% endblock %}
```

Jkrzy (changes requested):

crt_portal/cts_forms/templates/forms/base.html
cantsin: Oh gosh, egg on my face. Thanks for the catch!!

Jkrzy (approved):

Jkrzy:
@cantsin, this one failed to deploy to dev -- can you take a look?

PR #770, opened 2020-12-15
Filter reports by contact email

timoballard:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/820)

## What does this change?
Adds contact email as a report filtering option. All other filters remain active when an email filter is specified.

image

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1025,6 +1025,7 @@ class Filters(ModelForm):
             'hate_crime',
             'servicemember',
             'intake_format',
+            'contact_email',
         ]
 
         labels = {
@@ -1041,6 +1042,7 @@ class Filters(ModelForm):
             'violation_summary': 'Personal description',
             'create_date_start': 'Created Date Start',
             'create_date_end': 'Created Date End',
+            'contact_email': 'Contact email',
         }
 
         widgets = {
@@ -1080,6 +1082,12 @@ class Filters(ModelForm):
                 'name': 'violation_summary',
                 'placeholder': 'Personal Description',
                 'aria-label': 'Personal Description'
+            }),
+            'contact_email': EmailInput(attrs={
+                'class': 'usa-input',
+                'name': 'contact_email',
+                'placeholder': 'Contact Email',
+                'aria-label': 'Contact Email', 
             }),
         }
         error_messages = {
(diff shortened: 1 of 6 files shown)


crt_portal/cts_forms/tests/test_forms.py
```diff
@@ -361,6 +361,12 @@ def setUp(self):
         for index, report in enumerate(reports):
             report.assigned_section = sections[index % len(sections)]
             report.save()
+        # generate random reports associated with a different email address
+        reports = [Report.objects.create(**SAMPLE_REPORT) for _ in range(5)]
+        for report in reports:
+            report.assigned_section = 'VOT'
+            report.contact_email = 'SomeoneElse@usa.gov'
+            report.save()
```
Jkrzy, line 369: It looks like the new test here is the only one using these reports -- let's move their generation into that test to avoid creating these for every `FormNavigationTests` test.

crt_portal/cts_forms/forms.py
```diff
@@ -1081,6 +1083,12 @@ class Meta:
                 'placeholder': 'Personal Description',
                 'aria-label': 'Personal Description'
             }),
+            'contact_email': EmailInput(attrs={
+                'class': 'usa-input',
+                'name': 'contact_email',
+                'placeholder': 'Email',
```
Jkrzy, line 1089: Ahh missed this on first glance, let's have the placeholder be `Contact Email` to be consistent across the filter inputs.
```suggestion
                'placeholder': 'Contact Email',
```

Jkrzy (changes requested):
Looks :amaze: @timoballard, thanks! Made one suggestion in-line above.

I'd love to take this opportunity, or at least create a card to consider, to change how we're handling email searching on the backend as well. 

As-is we're doing a [full-text search](https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/search/) on the `contact_email` column but I think something like `__icontains` would be more appropriate. 

Email-like strings are going to be considered a single token by `to_tsvector` which means someone needs to enter the exact email they're trying to filter on. Searches for partial terms like "@gmail.com" will return no results.

timoballard:
100% agree that `__icontains` would be more appropriate here. Do you have thoughts on whether to make the change here or create a separate card for that? I lean slightly toward a separate card.

Jkrzy (approved):

PR #771, opened 2020-12-16
Add frequent flyer count to view all table

timoballard:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/819)

## What does this change?
Adds a column to the view all table that displays the total number of reports associated with each row's contact email. This PR does **NOT** include the functionality to sort by the total report count.

image

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/tests/test_forms.py
+++ b/crt_portal/cts_forms/tests/test_forms.py
@@ -760,4 +760,4 @@ class FiltersFormTests(TestCase):
             elif row['report'].contact_email == self.email2:
                 self.assertEqual(row['email_report_count'], 5)
             elif row['report'].contact_email is None:
-                self.assertEqual(row['email_report_count'], 8)
+                self.assertEqual(row['email_report_count'], None)
--- a/crt_portal/cts_forms/tests/test_forms.py
+++ b/crt_portal/cts_forms/tests/test_forms.py
@@ -737,7 +737,7 @@ class FiltersFormTests(TestCase):
         for report in reports_email_1:
             report.contact_email = self.email1
             report.save()
-        
+
         # generate reports for a different email address
         reports_email_2 = [Report.objects.create(**SAMPLE_REPORT) for _ in range(5)]
         for report in reports_email_2:
@@ -761,5 +761,3 @@ class FiltersFormTests(TestCase):
                 self.assertEqual(row['email_report_count'], 5)
             elif row['report'].contact_email is None:
                 self.assertEqual(row['email_report_count'], 8)
-
-
--- a/crt_portal/cts_forms/tests/test_forms.py
+++ b/crt_portal/cts_forms/tests/test_forms.py
@@ -722,3 +722,44 @@ class BulkActionsFormTests(TestCase):
         ]
         for action in form.get_actions(queryset.first()):
             self.assertTrue(action in expected_actions)
+
+
+class FiltersFormTests(TestCase):
+    def setUp(self):
+        self.client = Client()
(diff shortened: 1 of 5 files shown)


crt_portal/cts_forms/views.py
```diff
@@ -301,8 +301,11 @@ def index_view(request):
             p_class_list = p_class_list[:3]
             p_class_list[2] = f'{p_class_list[2]}...'
 
+        email_report_count = Report.objects.filter(contact_email=report.contact_email).count()
```
cantsin, line 304: Seems a bit slow to me to do this in a loop especially if the `per_page` QP is set to a high number (normally it defaults to 15).

Is the intention to get this in there and optimize later?

crt_portal/cts_forms/tests/test_forms.py
```diff
@@ -722,3 +722,42 @@ def test_bulk_actions_change_section_resets_user(self):
         ]
         for action in form.get_actions(queryset.first()):
             self.assertTrue(action in expected_actions)
+
+
+class FiltersFormTests(TestCase):
```
cantsin, line 727: Love the tests :heart:

crt_portal/cts_forms/views.py
cantsin: Also, I know you already spoke with Emily but are we going to adjust this PR to display `-` if there's no email or do this in a different PR?

cantsin (approved):
LGTM, a couple of questions

crt_portal/cts_forms/views.py
timoballard: Thanks for reviewing!

1. Agreed that this is not optimal. I fiddled a little but couldn't immediately figure out how to `aggregate/annotate` the `Report` query such that each record includes the total count of records with the same `contact_email`. The thinking here is that (at least) this part of the implementation will have to be refactored anyway when we incorporate sorting, which I presume will happen pretty much immediately after this gets merged. So this was more of a quick & dirty solution to at least get the counts displaying as a part of this sprint. 
2. Yes, I'm going to mark this PR as WIP until I hear from Emily about what she wants to do in the email=None case.

Jkrzy: The aggregation I landed on when fiddling with this a little bit was something like...count just the contact_emails we're about to render w/
```python
    emails_in_reports_on_this_page = [r.contact_email for r in requested_reports]
    emails_counts = Report.objects.filter(contact_email__in=emails_in_reports_on_this_page).values('contact_email').annotate(total=Count('contact_email'))
```

1 more query per page -Returns a queryset like which could be mapped back onto Reports as their being passed to context
```
<QuerySet [{'contact_email': 'testing@test.edu', 'total': 19}, {'contact_email': 'testing@test.test', 'total': 11}]>
```

crt_portal/cts_forms/views.py
```diff
@@ -287,6 +288,9 @@ def index_view(request):
 
     all_args_encoded = urllib.parse.quote(f'{page_args}&page={page}')
 
+    emails_in_reports_on_this_page = [r.contact_email for r in requested_reports]
```
cantsin, line 291: May be slightly more efficient to do `requested_reports.values_list('contact_email', flat=True)`? Wonder if this will help the ORM with optimizing the aggregation query.

cantsin (approved):
I had a couple of minor comments but after thinking about it, this PR looks good as is, let's get this in.

PR #773, opened 2020-12-17
Referred report flag

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/822)

## What does this change?

Adds a "Secondary Review" checkbox (nee "Referral") to report actions.

### Author

+ [x] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.).
+ [x] Check for, document, and establish a testing plan for any behavior that may vary across environments or is otherwise difficult to test.
+ [x] Check for [accessibility](/docs/a11y_plan.md).
+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

+ [x] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.).
+ [x] Check for any behavior that may vary across environments or is difficult to test, and ensure that it is well-understood, documented, and that there is a testing plan in place.
+ [x] Re-check for [accessibility](/docs/a11y_plan.md).
+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

## Notes for reviewer:

See [PR instructions doc](https://github.com/usdoj/crt-portal/blob/master/docs/pull_requests.md) for full pull request review instructions.

PR #776, opened 2020-12-18
Make reports sortable by email count

timoballard:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/843)

## What does this change?
Makes the report table sortable by the total number of reports associated with each report's contact email address.

image

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/views.py
```diff
@@ -255,11 +255,20 @@ def index_view(request):
...
-    requested_reports = report_query.order_by(*sort)
+    requested_reports = report_query.annotate(email_count=F('email_report_count__email_count'))
+
+    # apply the sort items individually so that we can push nulls to the back
+    for sort_item in sort:
+        nulls_last = 'email_count' in sort_item
+        if sort_item[0] == SORT_DESC_CHAR:
+            requested_reports = requested_reports.order_by(F(sort_item[1::]).desc(nulls_last=nulls_last))
+        else:
+            requested_reports = requested_reports.order_by(F(sort_item).asc(nulls_last=nulls_last))
```
timoballard, line 270: I don't _love_ this bit, so open to any clever alternative approaches y'all might come up with. The issue here is that Django's `nulls_last` arg has to be supplied to `asc` or `desc`, which means we have to call them explicitly rather than shorthanding (e.g. `-email_count`). So here we're unwinding the query params and pushing the sort fields into the query one by one to enable that.

crt_portal/cts_forms/tests/test_forms.py
```diff
@@ -745,19 +746,51 @@ def setUp(self):
             report.save()
 
         # generate reports with no email address
-        reports_email_none = [Report.objects.create(**SAMPLE_REPORT) for _ in range(8)]
+        reports_email_none = [Report.objects.create(**SAMPLE_REPORT) for _ in range(7)]
```
Jkrzy, line 749: This is starting to feel cumbersome, list comprehensions to generate sets of Reports and then iterating through them again to modify their values, flagging for us to address in another issue.

crt_portal/cts_forms/migrations/0093_auto_20201218_1557.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cts_forms', '0092_profile'),
+    ]
+
+    operations = [
+        migrations.RunSQL("""
+            CREATE OR REPLACE VIEW email_report_count AS
```
Jkrzy, line 14: Would it be easier to select `cts_forms_report.*` here instead of only ID/count and then joining them again with the columns from report?

crt_portal/cts_forms/views.py
Jkrzy: This doesn't handle multiple sort items since each order_by() call will clear any previous ordering. [ref](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#order-by)
As-is, `sort = ['status', '-email_count']` would only order by `email_count`

The current UI only allows folks to sort by a single field, let's handle just that case for now.

Jkrzy (changes requested):
Thanks @timoballard! Left a few comments in-line.

crt_portal/cts_forms/templates/forms/complaint_view/index/complaints_table.html
```diff
@@ -63,7 +63,7 @@
               </td>
               <td>
                 <a class="td-link display-block" href="{{datum.url}}">
-                  {% with count=datum.email_report_count|default:"—" %}
+                  {% with count=datum.report.email_count|default:"—" %}
```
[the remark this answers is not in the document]
Jkrzy: No -- since `email_count` is the annotation you're using below!

crt_portal/cts_forms/views.py
timoballard: 🤦‍♂️ right you are, thanks! I was thinking the `order_bys` stacked. This could be modified slightly to instead build up a list of query expressions to be supplied to `order_by` if/when we want to support multiple sort items. Perhaps we should do that now, since leaving this as it is makes it appear that we support it on the backend when in reality we don't

crt_portal/cts_forms/migrations/0093_auto_20201218_1557.py
timoballard: I went back and forth on this. The thing I was trying to avoid was having the `EmailReportCount` model duplicate all of the `Report` fields, because there are a lot of them. I thought about having `EmailReportCount` inherit from `Report`, but that seemed awkward. Do you have thoughts on this?

crt_portal/cts_forms/tests/test_forms.py
timoballard: https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/846

crt_portal/cts_forms/migrations/0093_auto_20201218_1557.py
Jkrzy: Sounds good to me, I think this is a reasonable trade-off for now! Thanks for considering different paths.

The Inheriting doesn't strike me as too awkward since I _think_ we can get by with something as small as:
```python
class EmailReportCount(Report):
    """
    Model for view created in `0093_create_email_report_count_view` where we append
    aggregate counts of `contact_email` to each `Report` record
    """  
    email_count = models.IntegerField()

    class Meta:
        managed = False
        db_table = 'email_report_count'
```

Jkrzy:
With the merge of https://github.com/usdoj-crt/crt-portal/pull/773 I think we'll need updated migrations here

crt_portal/cts_forms/views.py
timoballard: Going to fix this so that we actually support multiple sorts on the backend, stand by

crt_portal/cts_forms/migrations/0094_create_email_report_count_view.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.17 on 2020-12-18 20:57
```
cantsin, line 1: We merged in `0094_eeoc_referral_letters.py` as a migration recently so you'll have to bump up the numbers (and dependencies here). Sorry about that -_-

cantsin (changes requested):

cantsin (changes requested):
I'm getting a `FieldError at /form/view/106/`: `Cannot resolve keyword 'email_count' into field.` error when I do some advanced filtering. From the looks of it, I think `reconstruct_query` needs to know how to sort by the new `email_count` -- currently it cannot because the annotation is not present in that bit of the code. 

This happens when I have a `querydict` of 

```json
{"index": "3",
 "next": "?per_page=15&assigned_section=ELS&status=new&status=open&sort=-email_count&page=1"}
```

Could also be an error on my end, but I've ran the migrations and restarted docker compose. To replicate:

Go to the view all page, apply random filters, sort by `-email_count` and then click on any report.

timoballard:
thanks for keeping me honest @cantsin! 🤩 

Addressed the `reconstruct_query` problem, added some tests around it, and updated the nearby tests to use `ReportFactory`

cantsin (changes requested):
Looks good! One small bug: the index is incorrect when we sort by `email_count`. If I sort accordingly and then view the first report on the first page, I see `3 of 175 records` (when it should say `1 of ...`). 

From the looks of it, the code seems correct (and accounts for `email_count`) so I'm not sure what's happening here.

cantsin:
Also, thinking more broadly, I think we're seeing some unexpected interactions between sorting, filtering, pagination, and navigation. It shouldn't be this tricky to get right. (I take full responsibility, as I wrote some of this code :sweat_smile:) Maybe there's an opportunity here to rethink what the "right" approach should be here -- maybe use a library instead? We've had to patch all of this piecemeal, over time, and it's starting to creak at the edges a bit.

cantsin (approved):
Yay!! LGTM! :heart_eyes_cat:

PR #779, opened 2020-12-22
Render contact_email link on Report details page

Jkrzy:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/838)

## What does this change?
For each report details page we now render a `View All` link that shows a count of reports with identical `contact_email` values. The link is a pre-populated search on `contact_email` 

No link is shown if there are less than 2 reports found for a given `contact_email`.

Screen Shot 2020-12-22 at 3 50 17 PM

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/templates/forms/complaint_view/show/correspondent_info.html
```diff
@@ -26,7 +26,18 @@
       </tr>
       <tr>
         <th>Email</th>
-        <td>{{ data.contact_email|default:"—" }}</td>
+        <td>
+          <div class="display-flex">
+            <span>{{ data.contact_email|default:"—" }}</span>
+            {% if data.related_reports and data.related_reports|length > 1 %}
+              <span class="margin-left-auto">
+                <a href="{% url 'crt_forms:crt-forms-index' %}?contact_email={{data.contact_email}}">View All ({{data.related_reports|length}})</a>
```
cantsin, line 34: Wondering if we should list all sections explicitly in the query parameters, otherwise email matches from other sections may be hidden due to how we automagically filter for the user's given section in the view all page.

cantsin (changes requested):
One question related to section filtering otherwise LGTM

crt_portal/cts_forms/templates/forms/complaint_view/show/correspondent_info.html
cantsin: I realize what I said might not be clear, so here's a screenshot version:

Suppose I am an user with ELS, who sees this nifty feature. I see this report has one other email associated with it so I click on this nifty new feature:



But much to my surprise, only one shows up:



The "right" solution ime would be to _not_ filter for ELS anymore and then both reports would show up as advertised:

Jkrzy: Great question, bringing back to the group for discussion!!!

Our current implementation doesn't appear to let us conveniently override the "global section filter" 😅

Jkrzy: Revised this a bit to better handle this situation and reveal reports from all sections when following that link. Thanks @cantsin!

cantsin (changes requested):
I think we need to tweak the javascript as well when `assigned_section` is present as a query parameter, otherwise we get:

cantsin (changes requested):
Also, leaving a note to myself so I don't forget: might want to double check that the "print report" functionality does _not_ show the new link. (What'd be the point?)

Jkrzy:
Ready for another 👀 @cantsin! I'm not in love with this approach but it felt the least dramatic given the current state of `index_view`, the form classes and `Profile` concept, and how query parameters are being used across pages. 

Section filters, when provided as query parameters, now bypass any filters stored in a user's `Profile`. If no sections are provided as query parameters, the Profile is used. The section/profile form on the view all table always shows the filters which are currently applied, from the profile or request.

cantsin (approved):
Great work! :amaze: LGTM

PR #780, opened 2020-12-22
Add nonce to unsupported browser modal script

timoballard:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/823)

## What does this change?
Adds a nonce to the inline script that displays the unsupported browser modal at `/reports`

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,23 @@ install-cf7: &install-cf7
       rm cf7.deb
       cf7 api https://api.fr.cloud.gov
 
+install-python-dependencies: &install-python-dependencies
+  run:
+    name: Install python dependencies
+    command: |
+      pip install -U pip
+      pip install pipenv
+      pipenv install
+
+install-python-dev-dependencies: &install-python-dev-dependencies
+  run:
+    name: Install python dev dependencies
+    command: |
+      pip install -U pip
+      pip install pipenv
+      pipenv install --dev
+      pipenv run python -m playwright install
+
 version: 2
 jobs:
   build_and_test: # runs not using Workflows must have a `build` job as entry point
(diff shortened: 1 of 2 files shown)


Jkrzy:
This a great case to expand our playwright E2E tests to run against deployed environments, let's write a test confirming this works as expected when we're using `csp.middleware.CSPMiddleware`.

Running our tests against the development instance fails 
```bash
pipenv run pytest crt_portal/cts_forms/tests/integration/* --base-url=https://crt-portal-django-dev.app.cloud.gov
```

and the associated console error is visible in playwright with something like:

```python
@pytest.mark.only_browser("chromium")
def test_unsupported_browser_modal_visible(page):
    # print message received by console
    page.on('console', lambda x: print(x))
    page.goto("/report")
    page.waitForSelector("text=NOTICE: Your mobile browser is not compatible with this form",state='visible',timeout=3000)
```

Jkrzy (changes requested):
Let's cover this with a test

timoballard:
By "expand our playwright E2E tests to run against deployed environments", do you mean adding a post-deploy E2E test run to the circleCI config? So we'd continue doing what we're doing, but then after executing the deploy, run E2E again against the new instance?

.circleci/config.yml
```diff
@@ -191,6 +191,9 @@ jobs:
           name: Deploy to dev
           command: |
               ./blue-green-deploy.sh $CRT_USERNAME_DEV $CRT_PASSWORD_DEV dev
+      - run:
```
Jkrzy, line 194: I suspect this will fail since playwright and pytest are development dependencies and not installed here, we should be able to borrow the set-up steps from `build_and_test` above, bonus points if we can define it in a single place and re-use.

Jkrzy (changes requested):

Jkrzy (approved):
:amaze: Thanks Tim!  I re-targeted his against the staging/release candidate branch. We'll need to merge that back onto develop to catch these in our default branch.

PR #778, opened 2020-12-22
Add django-debug-toolbar

Jkrzy:
[Link to ZenHub issue.](link-goes-here)

## What does this change?
In reviewing https://github.com/usdoj-crt/crt-portal/pull/776 I used [django-debug-toolbar ](https://django-debug-toolbar.readthedocs.io/en/latest/index.html)to inspect the SQL being generated within`index_view`. It was super helpful and I think other devs would benefit from it's inclusion.

Making  available in local development environments to assist with code reviews, development, and troubleshooting.

Thought: Should this be more easily toggleable via an environment variable set in `docker-compose` ?

### Author

+ [ ] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.).
+ [x] Check for, document, and establish a testing plan for any behavior that may vary across environments or is otherwise difficult to test.
+ [x] Check for [accessibility](/docs/a11y_plan.md).
+ [ ] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

+ [ ] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.).
+ [ ] Check for any behavior that may vary across environments or is difficult to test, and ensure that it is well-understood, documented, and that there is a testing plan in place.
+ [ ] Re-check for [accessibility](/docs/a11y_plan.md).
+ [ ] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

## Notes for reviewer:

See [PR instructions doc](https://github.com/usdoj/crt-portal/blob/master/docs/pull_requests.md) for full pull request review instructions.

PR #789, opened 2020-12-28
Disable Django Debug Toolbar during test runs

timoballard:
## What does this change?
Related to #778 - Django test runs are run with `DEBUG=False`, regardless of what you specify in `settings`. This leads to url reversing issues when running tests locally, due to the fact that we toggle the django-debug-toolbar based on this setting. 

This is one option - disable the toolbar when executing tests. We could alternatively use a new setting (i.e. not `DEBUG`) to toggle the toolbar.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/crt_portal/settings.py
+++ b/crt_portal/crt_portal/settings.py
@@ -379,8 +379,7 @@ LOGGING = {
 if environment == 'LOCAL':
     from .local_settings import *  # noqa: F401,F403
 
-IS_RUNNING_TEST_SUITE = (os.path.basename(sys.argv[0]) == 'manage.py'
-                        and len(sys.argv) > 1 and sys.argv[1] == 'test')
+IS_RUNNING_TEST_SUITE = (os.path.basename(sys.argv[0]) == 'manage.py' and len(sys.argv) > 1 and sys.argv[1] == 'test')
 
 # Django debug toolbar setup
 if DEBUG and not IS_RUNNING_TEST_SUITE:
--- a/crt_portal/crt_portal/settings.py
+++ b/crt_portal/crt_portal/settings.py
@@ -12,7 +12,6 @@ https://docs.djangoproject.com/en/2.2/ref/settings/
 
 import json
 import os
-import sys
 
 import boto3
 from django.utils.log import DEFAULT_LOGGING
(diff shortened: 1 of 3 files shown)


crt_portal/crt_portal/settings.py
```diff
@@ -378,8 +379,10 @@
 if environment == 'LOCAL':
     from .local_settings import *  # noqa: F401,F403
 
+IS_RUNNING_TEST_SUITE = (os.path.basename(sys.argv[0]) == 'manage.py' and len(sys.argv) > 1 and sys.argv[1] == 'test')
```
cantsin, line 382: Hmm. This might not be needed. We already have similar (sigh, admittedly hacky) functionality for detecting environments. Not sure if this will work with CI, but maybe something like the following would work:

```python
if DEBUG and environment not in ['UNDEFINED', 'LOCAL']: ...
```

(Note: this approach is used elsewhere in this settings file -- I'm not a fan but if it works, let's strive for consistency?)

cantsin (changes requested):

crt_portal/crt_portal/settings.py
cantsin: Otherwise I would advocate for a separate setting configuration (if it's not too much work).

timoballard: Checking against the environment doesn't really help us in this case, if I understand things correctly, because at bottom, the issue is that...

https://github.com/usdoj-crt/crt-portal/blob/5dda152a2a1f8b79a5f84a1202aae5980486b57b/crt_portal/crt_portal/urls.py#L90-L93

... will resolve to `False` when running `manage.py test`, and `True` when running `manage.py runserver`, all other things being equal. 

That said, after some more thought I also think it would be cleaner to capture this toggling as its own configuration, so I'll push that change momentarily. Only thing to keep in mind with this approach is that the toolbar will (by default) be enabled for local test runs, which maybe isn't a big deal

cantsin: Ahh I see now. Yes, thank you for the explanation -- I wasn't looking closely enough at this issue. What you have LGTM.

cantsin (approved):
Thanks for the config change! Cleaner now :star_struck:

PR #794, opened 2021-01-05
Compressor redux

cantsin:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/792)

## What does this change?

Second try at implementing django-compressor functionality. Believe it or not, the `hero.png` image (which we never use) is indeed needed and breaks [manifest storage](https://docs.djangoproject.com/en/2.2/ref/contrib/staticfiles/#manifeststaticfilesstorage) outright if we omit this image. Why? Well, the USWDS CSS specifies `background-image:url("../img/hero.png");` which the Django manifest storage backend picks up and then subsequently crashes during `collectstatic` when this file is not found. tl;dr: I believe this missing, unused file is what broke deployment last time (really). 

The rest is identical to before, except for a couple of small things. We can use the built in `S3ManifestStaticStorage` instead of monkey patching classes. Two CSS `url` references have been tweaked so that they are properly fetched from S3. 

But most exciting is the `localstack` documentation I've thrown in, which explains how to set up an identical-ish production environment locally.

### Author

### Reviewer

## Notes for reviewer:


cantsin:
@Jkrzy flagging you as a reviewer too -- oddly enough Github is not populating your username in the reviewers field for whatever reason.

cantsin:
Anyway, I'd like one of y'all to go through the documentation and let me know what worked, what was unclear, what could be improved. Thanks!

Jkrzy:
This is :amaze:ing @cantsin!!! 

Running through the docs I get tripped up on `collectstatic`  with:

`s3_1   | 2021-01-07T15:23:09:INFO:localstack.services.edge: Unable to find forwarding rule for host "s3:4566"`

and this exception from boto: `botocore.exceptions.ClientError: An error occurred (404) when calling the PutObject operation: Not Found`

Did you hit this in local testing?

cantsin:
@Jkrzy I have not gotten that error! To double check the obvious, did you rebuild the docker-compose image? I just ran through the steps again myself and so far I don't see that error. 

- Does `http://localhost:4566/health?reload` show the s3 service running? 
- If you do `docker-compose run web curl http://s3:4566/health` what shows up?

These two test external and internal access to the localstack edge port, respectively.

cantsin:
@Jkrzy updated to address the issue we found last Friday. Please note that you will need to remove your local S3 bucket and re-create it (since `collectstatic` will skip if the file is already on S3 and therefore not save to `staticfiles` in the container as per our new storage class). Alternatively, I think removing the files: `aws s3 rm --recursive s3://crt-portal/ --endpoint-url=http://localhost:4566` will also work.

Tested this time by doing `git clean -fxd` beforehand :grinning:

crt_portal/crt_portal/storage.py
```diff
@@ -0,0 +1,16 @@
+from django.core.files.storage import get_storage_class
+from storages.backends.s3boto3 import S3Boto3Storage
+
+
+class CachedS3Boto3Storage(S3Boto3Storage):
```
Jkrzy, line 5: I think we want `S3ManifestStaticStorage` to keep the file hashing/cache breaking.
```suggestion
from storages.backends.s3boto3 import S3ManifestStaticStorage


class CachedS3Boto3Storage(S3ManifestStaticStorage):
```

crt_portal/crt_portal/storage.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    Save compressed files to local storage before uploading to S3.
+    """
+    def __init__(self, *args, **kwargs):
+        super(CachedS3Boto3Storage, self).__init__(*args, **kwargs)
+        self.local_storage = get_storage_class("compressor.storage.CompressorFileStorage")()
+
+    def save(self, name, content):
+        self.local_storage._save(name, content)
+        super(CachedS3Boto3Storage, self).save(name, self.local_storage._open(name))
```
Jkrzy, line 15: Quick fix to update this usage of super
```suggestion
        super().__init__(*args, **kwargs)
        self.local_storage = get_storage_class("compressor.storage.CompressorFileStorage")()

    def save(self, name, content):
        self.local_storage._save(name, content)
        super().save(name, self.local_storage._open(name))
```

Jkrzy (changes requested):
Thanks @cantsin! We're super close but this latest change reverts to the non-manifesting storage implementation.

crt_portal/crt_portal/storage.py
cantsin: Thanks for both suggestions! I had `S3ManifestStaticStorage` locally but did not commit it :man_facepalming: Going through the paces once more to make sure everything is kosher.

Jkrzy (approved):
Looks good, thanks for sticking with this one @cantsin!

PR #807, opened 2021-01-28
Add secondary review filter

timoballard:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/870)

## What does this change?
Adds a Secondary Review report filter. 

image

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/tests/test_forms.py
+++ b/crt_portal/cts_forms/tests/test_forms.py
@@ -944,6 +944,7 @@ class FiltersFormTests(TestCase):
         for row in response.context['data_dict']:
             self.assertTrue(row['report'].referred)
 
+
 class SimpleFilterFormTests(SimpleTestCase):
 
     def test_get_sections_returns_only_valid_choices(self):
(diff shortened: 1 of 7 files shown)


crt_portal/cts_forms/tests/test_forms.py
```diff
@@ -932,6 +932,18 @@ def test_email_count_caseinsensitive(self):
             if row['report'].contact_email in [self.email4, self.email5, self.email6, self.email7]:
                 self.assertEqual(row['report'].email_count, 15)
 
+    def test_secondary_review_filter(self):
+        ReportFactory.create_batch(5, referred=True)
+
```
Jkrzy, line 937: As an integration test this would cover better if we had both referred and non-referred reports.  As-is this passes if `referred` filtering isn't implemented.

```suggestion
        ReportFactory.create(referred=True)
        ReportFactory.create(referred=False)
```

Jkrzy (changes requested):
One small test tweak, looks good otherwise! Thanks @timoballard

crt_portal/cts_forms/tests/test_forms.py
Jkrzy: Whoops, there are already `16` reports created in setUp before these 5. So this does fail as-is without the filtering. 

I retract my suggestion and re-enter as broader thought -  "We're creating an awful lot of reports in testing" 😅

Jkrzy (approved):
On a [second look at the test in question](https://github.com/usdoj-crt/crt-portal/pull/807/files#r566896300), this is OK as-is.

PR #820, opened 2021-03-01
Il/attachments download

ilodidoj:
[Link to ZenHub issue.](https://app.zenhub.com/workspaces/doj-crt-intake-scrum-board-5d03bf56c1c8a35d482eca0f/issues/18f/crt-portal/884)

## What does this change?
Allow user to download attachment from a report

### Author

### Reviewer

## Notes for reviewer:


ilodidoj:
@Jkrzy Tim and I paired on the download.  We can still get some of your thoughts, please review. CC: @timoballard

Jkrzy:
Thanks @ilodidoj! Can you say more in the PR about _how_ this approach addresses the issue? I'd love to hear more about this approach and alternatives that ya'll considered.

ilodidoj:
@Jkrzy thanks for looking into it! We have two alternatives for this issue, one is to render the attachment directly using Django and secondly via AWS S3.  Our collective thought on this attachment download mostly be used by the intake specialist. Having downloading big files by multiple intake specialists would create resource burden and degraded user experience (longer time on the modal downloading... ) as oppose to AWS S3 direct download.  Considering that we went the route of presigned URL from S3 aka second option.  This may show the AWS direct link for the file but only to the internal intake users with very limited burden on the portal app resources.

crt_portal/cts_forms/templates/forms/complaint_view/show/attachments.html
```diff
@@ -14,11 +14,19 @@
     <input type="hidden" value="{{ return_url_args }}" name="next" id="next" />
     <input type="hidden" value="{{ index }}" name="index" id="index" />
     <input type="hidden" value="{{ data.pk }}" name="report" id="report_id" />
-    {% for attachment in attachments %}
-      <p>{{ attachment.filename }}</p>
-    {% endfor %}
+        <table class="usa-table usa-table--borderless related-reports width-full">
```
Jkrzy, line 17: We should include headers describing the content of each cell when using tables to display data. Headers let folks know what to expect in a given cell and allow screen readers to provide context about the data as folks navigate the table.

Jkrzy:
Thanks for the details @ilodidoj. 

I'm a bit worried about this approach creating signed urls on each page render that are valid for an hour. It's extra work to generate the signed urls for every attachment on every template render and it allows folks to access files without us keeping an application level log entry.

What if each request to download an attachment was handled by the application, ensuring authentication, logging the access, and redirecting to a short-lived signed url?

ilodidoj:
@Jkrzy thanks for the good feedback. Actually we like your idea and it's not that big of a change but add good value. We are making this changes and pushing them as soon we finish.

crt_portal/cts_forms/templates/forms/complaint_view/show/attachments.html
ilodidoj: Great catch!! we talked about this earlier for table accessibility (508).  Sorry forgot to add that, should be in our next push. thanks.

crt_portal/cts_forms/models.py
```diff
@@ -365,6 +365,10 @@ class ReportAttachment(models.Model):
     report = models.ForeignKey(Report, on_delete=models.CASCADE, related_name='attachments')
     created_date = models.DateTimeField(auto_now_add=True)
 
+    @property
+    def download_url(self):
+        return self.file.name
+
```
Jkrzy, line 371: Good opportunity here to should drop this snippet and replace with a [get_absolute_url](https://docs.djangoproject.com/en/2.2/ref/models/instances/) method. 

It's great to use `get_absolute_url` or the [url templatetag](https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#url) instead of hard-coding.

crt_portal/cts_forms/views.py
```diff
@@ -635,10 +640,52 @@ def post(self, request):
...
+                response = s3_client.generate_presigned_url('get_object',
+                                                            Params={'Bucket': settings.PRIV_S3_BUCKET,
+                                                                    'Key': attachment.file.name,
+                                                                    'ResponseContentDisposition': f'attachment;filename={attachment.filename}'},
+                                                            ExpiresIn=30)
+
+                return redirect(response)
+
+            except ClientError as e:
+                logging.error(e)
```
Jkrzy, line 687: Would be great to send this to NewRelic in a later iteration.

crt_portal/cts_forms/views.py
```diff
@@ -635,10 +640,52 @@ def post(self, request):
...
+    http_method_names = ['get', 'post']
+
+    def get(self, request, id, filename):
+        """
+        Download a particular attachment for a report
+        """
+        report = get_object_or_404(Report.objects.prefetch_related('attachments'), pk=id)
+        attachment = report.attachments.get(file=f'attachments/{filename}')
+
+        logger.info(f'User {request.user} downloading attachment {attachment.filename} ({attachment.file.name}) for report {id}')
```
Jkrzy, line 655: I got caught up a bit here untangling the difference between, `filename`, `attachment.filename`, and `attachment.file.name`. 

Is there a way we can disambiguate these now before they become an acceptable pattern for us in the future?

Using `filename` in the URL doesn't gain us much. Accessing these as `attachments/{attachment.id}` could be a bit simpler to handle.

crt_portal/cts_forms/templates/forms/complaint_view/show/attachments.html
```diff
@@ -14,11 +14,25 @@
...
+            <tbody>
+                    <tr>
+                        <th style="text-align:left">File Name</th>
+                        <th style="text-align:left">ID</th>
+                        <th style="text-align:left">Added By</th>
+                        <th style="text-align:left">Created Date</th>
+                    </tr>
```
Jkrzy, line 24: We're not working from mock-ups but having this inside of <tbody> is rendering a top-border that looks like a duplicative. 
Let's move it into <thead>, remove the inline styling, and apply scope='col'.

Screen Shot 2021-03-03 at 9 24 26 AM


I'm also in favor of dropping  the `ID` column, it's going to be identical for every row in the table.

Jkrzy (changes requested):
Looking great, thanks @ilodidoj! Left a few comments inline, I'd love to tackle the table headers and `filename` disambiguation in `ReportAttachmentView.get` before merging this in.

crt_portal/cts_forms/models.py
ilodidoj: we applied get_absolute_url method. Thanks for the tips.

crt_portal/cts_forms/views.py
ilodidoj: Agreed, we can explore in a different card.

crt_portal/cts_forms/views.py
ilodidoj: simplified as per suggestions.

crt_portal/cts_forms/templates/forms/complaint_view/show/attachments.html
ilodidoj: @Jkrzy We have made the necessary changes, please have another look. Thanks for all the input.  BTW, changing to get_absolute_url and filename to id cause test to fail.  That a small change we will update unit test soon.

crt_portal/cts_forms/models.py
```diff
@@ -365,6 +365,9 @@ class ReportAttachment(models.Model):
     report = models.ForeignKey(Report, on_delete=models.CASCADE, related_name='attachments')
     created_date = models.DateTimeField(auto_now_add=True)
 
+    def get_absolute_url(self):
+        return f'attachments/{self.id}'
```
Jkrzy, line 369: We should be using [reverse](https://docs.djangoproject.com/en/2.2/ref/urlresolvers/#reverse) when generating URLs. As-is this is going to fail if/when we want to use it _anywhere_ else besides the report details page.

```suggestion
        return reverse('crt_forms:get-report-attachment', kwargs={"id": self.report.id, "attachment_id": self.id})
```

Jkrzy:
Looks great! We'll do a final pass once the tests are passing.

crt_portal/cts_forms/views.py
Jkrzy: Since we have `attachment_id` in the url now we should get this directly by attachment ID

timoballard: hm, I'm seeing ☝️ this `attachment_id` change - GitHub might be showing an outdated version @Jkrzy, which happens to me all too frequently

Jkrzy: :woah: You're right, my mistake. Thanks!

Jkrzy (approved):
Thanks for fixing up the tests!

PR #914, opened 2021-07-20
Preloading hero images for better loading time on the homepage

LindsayYoung:
This was a small think I stumbled on. Saves .7 seconds on page load on landing page. 

We should look at fonts for further optimization.


ilodidoj (approved):
LGTM. Local flake8 and unit test passed. Working as intended.

PR #938, opened 2021-08-19
Add safeguards to prevent javascript error

billyhunt:
[Link to ZenHub issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1030)

## What does this change?

The close button on the trafficking modal is not working.  There is an error that happens, shown below.  This PR fixes the error by not scrolling when the form does not allow it.  

image
image

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/templates/forms/report_base.html
+++ b/crt_portal/cts_forms/templates/forms/report_base.html
@@ -137,7 +137,7 @@
       root.CRT.openModal(unsupportedBrowserModal);
 
       var close_el = document.getElementById('ub_close');
-      root.CRT.cancelModal(modal, close_el);
+      root.CRT.cancelModal(unsupportedBrowserModal, close_el);
     }
   })(window);
 </script>
--- a/crt_portal/static/js/modal.js
+++ b/crt_portal/static/js/modal.js
@@ -57,7 +57,7 @@
 
   root.CRT.cancelModal = function(modal_el, cancel_el, form_el = {}) {
     var dismissModal = function(event) {
-      if (form_el) {
+      if (form_el && form_el.scrollIntoView && form_el.focus) {
         form_el.scrollIntoView({ behavior: 'smooth', block: 'end', inline: 'nearest' });
         form_el.focus();
       }
--- a/crt_portal/static/js/modal.js
+++ b/crt_portal/static/js/modal.js
@@ -55,9 +55,9 @@
     dom.body.classList.remove('is-modal');
   };
 
-  root.CRT.cancelModal = function(modal_el, cancel_el, form_el = {}) {
+  root.CRT.cancelModal = function(modal_el, cancel_el, form_el) {
     var dismissModal = function(event) {
-      if (form_el && form_el.scrollIntoView && form_el.focus) {
+      if (form_el) {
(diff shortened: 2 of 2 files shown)


louh:
So this does fix the issue, however, I wonder if it's worth considering correcting the logic in the function arguments itself.

On [line 58 of `modal.js`](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/modal.js#L58):

```js
 root.CRT.cancelModal = function(modal_el, cancel_el, form_el = {}) {
```

When `form_el` is not passed to the `cancelModal()` method (which is [the case here](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/redirect-modal.js#L34)) it is assigned a default value of empty object.

The original [`if` statement](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/modal.js#L60) checks to see if `form_el` is a falsy value:

```js
if (form_el) {
  // Do stuff
}
```

However, an empty object is _not_ falsy. This is why the `if` statement is always true, and this block will always run.

Rather than check to see if the object is not empty (and has the methods it wants to run) I believe we might be able to fix the problem by not assigning a default value to `form_el` at all. If `form_el` is undefined, it will pass the falsy check and it should avoid running the `if` block.

The benefit of doing this is some cleaner code, and the conditions on the `if` statement won't need to be updated with the behavior inside its block in the future.

billyhunt:
> So this does fix the issue, however, I wonder if it's worth considering correcting the logic in the function arguments itself.
> 
> On [line 58 of `modal.js`](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/modal.js#L58):
> 
> ```js
>  root.CRT.cancelModal = function(modal_el, cancel_el, form_el = {}) {
> ```
> 
> When `form_el` is not passed to the `cancelModal()` method (which is [the case here](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/redirect-modal.js#L34)) it is assigned a default value of empty object.
> 
> The original [`if` statement](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/modal.js#L60) checks to see if `form_el` is a falsy value:
> 
> ```js
> if (form_el) {
>   // Do stuff
> }
> ```
> 
> However, an empty object is _not_ falsy. This is why the `if` statement is always true, and this block will always run.
> 
> Rather than check to see if the object is not empty (and has the methods it wants to run) I believe we might be able to fix the problem by not assigning a default value to `form_el` at all. If `form_el` is undefined, it will pass the falsy check and it should avoid running the `if` block.
> 
> The benefit of doing this is some cleaner code, and the conditions on the `if` statement won't need to be updated with the behavior inside its block in the future.

That makes a lot of sense.  I will give it a go.

billyhunt:
I updated the implementation to @louh's  suggestion, tested all the modals, and fixed the unsupported browser modal, which had a different bug from last year. (good idea to check them all @ilodidoj)

louh (approved):
LGTM!

ilodidoj (commented):
@billyhunt I'll be merging this. Please also create one PR for the Release 08-24-2021 merge.

PR #971, opened 2021-09-08
Send auto-confirmation email on form submit

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1042)

## What does this change?

This is the "first slice" implementation of sending auto-response e-mails to public users who submit a report with an e-mail address contact.

Follow up qs

- [x] Do we need to log this in the activity log? If so, who is the user? (Update: ~~yes; should be "system"; need to know if it can be a string or does it need to match an entry in the model~~ no, we can only log in server output)
- [x] E2E tests are broken (see CircleCI). Needs investigation (Update: front-end throws an error if an email template is not found. This is fixed.)
- [x] Validate email before sending (Update: email is validated in the form itself, no need to do it in the email sending method)

### Author

### Reviewer

## Notes for reviewer:


louh:
Thanks for adding the template migration @billyhunt. This fixed the E2E tests because the .get() query to find the template will throw an error if it doesn't find it (or it finds more than one record). I updated the query so it's guaranteed to return one template or `None`, and we shouldn't see errors of this kind on the front-end anymore in case anything messes with the templates.

billyhunt:
sounds good.  I tested it on my side with no template and the correct one, and both acted as they should. Great job on this @louh

billyhunt (approved):
I think this looks really good.  We might want a followup ticket to add a test or two.

louh:
I'm recording here the summary of some internal conversations we've had to keep this bounded to a "thin slice" implementation.

**Activity logging**

Currently, emails sent through TMS are logged in the project's "actstream" as well as a record in the "TMS email" model. However, because this is an automated response, no authenticated user is initiating the email, therefore, we are unable to log it into actstream, whose "actor" field requires a reference to a User record. As a result, we will only log the automated response in server logs (with `logger.info()`, and a TMS email record will also exist, which is part of our default behavior.)

The trade-off with this resolution is that automated responses do not automatically show up in the report view's activity log. This requires effort to think through and potentially implement, and we should carve this out to consider in a future follow-up task.

**Tests**

As usual, new functionality should be adequately covered by tests. However, when we dug into the state of existing tests, there are a number of things that we should also consider carving out into a follow-up task:

- We do not have tests on the e-mail sending functionality right now. Writing those tests would be out of scope for this feature's thin slice, but we could take time to cover those later.
- Another approach to testing would be to do it in the integration / E2E testing, which runs through the entire steps of the form. However, this is currently front-end only and do not seem to have a way to test back-end behavior. Looking into this may also be out of scope for a "thin slice".

louh:
This PR also introduces the `EMAIL_AUTORESPONSE_ENABLED` environment variable. If present and set to true, an autoresponse email will be sent when a report contains a valid contact email. (This can be further controlled with allowed email addresses recipients.) In local environments, the default value is `True`. In all other environments, the default value of this variable is `False`. e.g. to enable email autoresponse in dev, set the `EMAIL_AUTORESPONSE_ENABLED` environment variable in that instance to`True`.

ilodidoj (approved):
LGTM

PR #985, opened 2021-09-22
fix: when closing a report, don't check against cached status

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1075)

## What does this change?

Potential fix for a bug where reports that are closed do not update its "closed date" property.

We think the issue is with a stale cached value. When [the `report.closed` property is checked](https://github.com/usdoj-crt/crt-portal/blob/4bdcbb4a4c05eb6b983a09d15210422d94150a2a/crt_portal/cts_forms/forms.py#L1349), it is not always true when it should be. This is because it's a [computed property that is cached](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/models.py#L356-L358).

The quick-fix is to stop checking against the cached property in this method.

Since we can't consistently cause a closed report to fail to update its closed date, we may need to run this for a while to ensure that we've solved the problem and that it doesn't create unnecessary performance bottlenecks.

n/a

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -15,7 +15,8 @@ from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from django.conf import settings
 
-from .model_variables import (COMMERCIAL_OR_PUBLIC_ERROR,
+from .model_variables import (CLOSED_STATUS,
+                              COMMERCIAL_OR_PUBLIC_ERROR,
                               COMMERCIAL_OR_PUBLIC_PLACE_CHOICES,
                               COMMERCIAL_OR_PUBLIC_PLACE_HELP_TEXT,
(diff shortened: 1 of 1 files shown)


louh:
Light bulb moment this morning: we could also just remove the `@cached_property` decorator (https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/cts_forms/models.py#L356), right? The benefit of this is `report.closed` is easier to read, and we won't need to import the `CLOSED_STATUS` constant into a module that doesn't otherwise need it.

louh:
So I like this change better. If it works, it's a one-line fix!

billyhunt (approved):
Looks good to me!

ilodidoj (approved):

PR #997, opened 2021-09-30
Bh/minify sourcemaps javascript

billyhunt:
[Link to ZenHub issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1084)

## What does this change?

We have minified the javascript, and added sourcemaps for debugging.  

### Author

### Reviewer

## Notes for reviewer:


gulpfile.js
```diff
@@ -17,6 +17,8 @@ const gulp = require("gulp");
 const pkg = require("./node_modules/uswds/package.json");
 const postcss = require("gulp-postcss");
 const replace = require("gulp-replace");
+const rename = require("gulp-rename")
+const uglify = require('gulp-uglify');
```
louh, line 21: Style nits:
- trailing semicolon on line 20
- use double quotes on line 21

gulpfile.js
```diff
@@ -43,6 +45,7 @@ const FONTS_DEST = "./crt_portal/static/fonts";
 
 // Javascript destination
 const JS_DEST = "./crt_portal/static/js";
+const JS_FILES = [`${JS_DEST}/*.js`, `!${JS_DEST}/*.min.js`]
```
louh, line 48: semicolon

gulpfile.js
```diff
@@ -76,6 +79,19 @@ gulp.task("copy-uswds-js", () => {
...
+gulp.task('build-js', function() {
+  return gulp.src(JS_FILES)
+    .pipe(sourcemaps.init())
+    // Minify the file
+    .pipe(uglify())
+    .pipe(rename({
+      suffix: '.min'
+    }))
+    .pipe(sourcemaps.write('.'))
+    // Output
```
louh, line 91: use double quotes throughout (we're copying the style from uswds-gulp, it's not based on any other intentional decision making on our end)

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "sass:watch": "gulp watch",
+    "gulp:watch": "gulp-watch",
```
louh, line 8: This doesn't work! Is the intention to split up sass and JS watching?

With this `npm run sass:watch` will also watch and build JS, but that's probably not what we want?

louh (changes requested):
This is great (minified JS is a necessary step toward better performance) but I had some JS style nits and also a request to take a closer look at the npm run scripts.

package.json
billyhunt: I fixed the bug, and changed sass:watch to only watch the sass files, and made gulp-watch to do watch both js and sass.

I also updated run.sh and config.yml to build the js in addition to the sass.

louh: If we're building JS on docker run and deploys, we don't need to check in the minified JS, right? This would be similar to how we don't currently check in the contents of `static/css/compiled` to source control.

billyhunt: That tracks.  I will update.

billyhunt: Updated.

.gitignore
```diff
@@ -21,3 +21,4 @@ crt_portal/NO_ZIP.txt
 *.mo
 # this is where we store uploaded attachments when USE_LOCALSTACK=False
 attachments/
+*.min.js*
```
louh, line 24: This is now too broad since we had some existing minified JavaScript that wasn't part of the build process (they were just manually stored in the repo):

- crt_portal/static/js/aria-autocomplete-1.2.3.min.js 
- crt_portal/static/js/uswds.min.js
- crt_portal/static/js/uswds.min.js.map

louh (changes requested):

.gitignore
louh: Oh, the last two will just get rebuilt since we also copied the source `uswds.js` in, so this might only affect the `aria-autocomplete-1.2.3.min.js` script.

billyhunt: I moved aria-autocomplete-1.2.3.min.js into another folder so it will not be built.  Everything else should build as we want.

louh: Cool, I like this approach. Is `precompiled` something you've seen as a folder name? Usually I see these named as `vendor` (e.g. [1](https://stackoverflow.com/questions/50052846/use-of-vendor-folder-in-web-design) and [2](https://tommcfarlin.com/the-vendor-directory/)). Happy to go with your recommendation but I wanted to suggest an alternative name that seems more common in my experience.

billyhunt: vendor works for me.  I will update.

billyhunt: updated

louh (approved):

PR #1027, opened 2021-11-03
Add filter for unassigned reports

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1116)

## What does this change?

Adds an "(unassigned)" option to the Assignee filter. The "(unassigned)" option is visible at the top of the list and can be reached via typeahead. If "(unassigned)" is selected, and the "Apply filters" button is clicked, reports without assignees will be shown in the resulting table.

Furthermore, I took the opportunity to replace the underlying third-party autocomplete component with the USWDS [combo box component](https://designsystem.digital.gov/components/combo-box/), which was first introduced in [v2.7.0](https://github.com/uswds/uswds/releases/tag/v2.7.0). Note that this only occurs here in the Assignee filter and not elsewhere that the assignee dropdown occurs; this can be resolved in follow up tickets, such as the one that addresses https://github.com/usdoj-crt/crt-portal-management/issues/739. The eventual, preferred outcome is to fully replace the third-party autocomplete library. Using the USWDS's built-in component is advantageous for a number of reasons:

1. It unifies the appearance of the typeahead dropdown with the rest of the rest of the app's UI.
2. It reduces the loading time needed, which currently makes an additional request for the third-party component.
3. We use a component that is developed in accordance with federal web accessibility guidance.

## Screenshots (for front-end PR):

Screen Shot 2021-11-03 at 10 14 31 AM

### Author

+ [x] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.).
+ [x] Check for, document, and establish a testing plan for any behavior that may vary across environments or is otherwise difficult to test.
+ [x] Check for [accessibility](https://github.com/usdoj-crt/crt-portal/blob/develop/docs/a11y_plan.md).
+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

### Reviewer

+ [x] If this is a story, run locally and check to make sure all Acceptance Criteria are met. (If any criteria are unclear, ask about them.).
+ [x] Check for any behavior that may vary across environments or is difficult to test, and ensure that it is well-understood, documented, and that there is a testing plan in place.
+ [x] Re-check for [accessibility](https://github.com/usdoj-crt/crt-portal/blob/develop/docs/a11y_plan.md).
+ [x] [Tests pass](https://github.com/USDOJ/crt-portal/#tests).

## Notes for reviewer:

[...]

PR #1030, opened 2021-11-08
Unassign from report actions (first pass)

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/739)

## What does this change?

We replace the old autocomplete widget in the "assign to" action with the new [combo box](https://designsystem.digital.gov/components/combo-box/) component from USWDS. There is some rationale about this move in [this previous task where the report filters can now filter for unassigned reports](https://github.com/usdoj-crt/crt-portal/pull/1027).

The new component has a "clear this value" button in its UI that the previous autocomplete widget did not have. The "clear this value" button only appears when the combo box value has a selected value. For a user to unassign a report, they only need to clear this value; this was [previously verified to be manually achievable by pressing backspace or deleting the contents of the input box](https://github.com/usdoj-crt/crt-portal-management/issues/739). However, this interaction was unclear to end users, hence there was an ask for a way to unassign the report in the UI. With a "clear this value" button, this need can be addressed using UI we already have. (See screenshots section below for a brief animation of this behavior.)

The alternative proposal (not implemented here) is to include a "unassign" option in the dropdown itself. This was how we implemented the unassign filter for https://github.com/usdoj-crt/crt-portal-management/issues/1116 (#1027), and I briefly explored doing the same in this case. However, the solution was less simple:

- Clicking the dropdown to select "unassign" required multiple clicks as opposed to clicking on the "clear this" X once. (If we determine later that users still struggle to understand the X, and prefer selecting something from a dropdown, then the multiple clicks may still be justified.)
- We have the "clear this" UI for free.
- Implementing a special option for "unassign" requires more work (for future note, though, don't re-implement the same widget used in the filter, since the filter needs to distinguish between "empty" and "unassigned," for this UI, "empty" == "unassign".)

Furthermore, I added tests in this PR and fixed a small issue that prevented the entire report action test suite from running in the past.

Nov-08-2021 12-44-42

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/forms.py
+++ b/crt_portal/cts_forms/forms.py
@@ -1214,15 +1214,18 @@ class ComplaintActions(ModelForm, ActivityStreamUpdater):
     assigned_to = ModelChoiceField(
         queryset=User.objects.filter(is_active=True).order_by('username'),
         # crt view only
-        label="Assigned to",
-        required=False
+        label='Assigned to',
+        required=False,
+        widget=Select(attrs={
+            'class': 'usa-input usa-select',
+        })
     )
     referred = BooleanField(
+        label='Secondary review',
         required=False,
-        label='Secondary Review',
         widget=CheckboxInput(attrs={
             'class': 'usa-checkbox__input',
-            'aria-label': 'Secondary Review',
+            'aria-label': 'Secondary review',
         })
     )
 
@@ -1272,7 +1275,6 @@ class ComplaintActions(ModelForm, ActivityStreamUpdater):
             choices=_add_empty_choice(DISTRICT_CHOICES, default_string=''),
             required=False
         )
-        self.fields['assigned_to'].widget.label = 'Assigned to'
 
     def get_actions(self):
         """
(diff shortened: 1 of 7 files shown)


billyhunt:
Great work!  The only think that would be a nice to have, is if you could delete the field, hit enter, and have it do the same behavior as if you hit the 'X'.  Currently, if you delete the field, it defaults back to the original entry.  

https://user-images.githubusercontent.com/6232068/140819591-c17afa07-e1b7-4e61-8f41-a2173b70df0f.mp4

louh:
@billyhunt I thought this was easier and it wasn't. It turns out that the USWDS combo box component hides the empty option that I needed to be there. I'm going to spend a bit of time today investigating how possible it is to patch a USWDS component.

billyhunt (approved):

billyhunt:
It might not be worth it if its going to cause up problems down the line.  Since this is an internal tool, I think its fine as is.

louh:
As a followup I'd opened a conversation over at the USWDS repo to see if some of this behavior is intentional for accessibility / usability reasons we're overlooking, but if not, I'd hope to propose some improvements: https://github.com/uswds/uswds/discussions/4391

In the meantime, I thought I'd add a "(nobody)" placeholder which might make it more clear what an empty field means.
Nov-09-2021 11-53-44

PR #1031, opened 2021-11-08
'Time to review' tracking in Activity Log

billyhunt:
[Link to ZenHub issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1125)

## What does this change?

The GAO (Government Accountability Office) would like to be able to track metrics on the time it takes from when a report is submitted, to when it is first read.  

Here is the workflow.  
1. When an Intake Specialist visits the detail view for a given report, the server will check to see if that page has been previously opened.
2. If opened = True, it does nothing.
3. If opened = False, then it will write a message to the activity log.  It also sets open = True, so we only write to the log one time.

Note: We had discussed not adding a DB entry for "opened", but after seeing the performance hit of searching through activity stream messages in Python, I thought it best to do the operation with a quick DB action. 

Report Detail View
Screen Shot 2021-11-08 at 12 37 30 PM

Report Admin Page
image

Actions Admin Page
image

### Author

### Reviewer

## Notes for reviewer:


louh (approved):
LGTM! This is super clean and efficient, and I agree with the rationale for adding the "opened" boolean property on the report model.

On the process side, do we need to do anything else to communicate to staff that the "opened" property will be inaccurate for any report submitted before this feature is launched?

ilodidoj (approved):

PR #1036, opened 2021-11-10
fix(combo-box): patch for z-index conflict

louh:
## What does this change?

Bug fix (reported by @billyhunt) for an issue with the combo box component where its child UI elements can appear on top of other site UI, e.g. in this case where the "Assigned to" select UI shows above the "Contact complainant" modal.

Screen Shot 2021-11-10 at 11 42 07 AM

The fix is to apply a z-index to the parent `.usa--combo-box` element which creates a local "stacking context" that keeps the child elements from leaking into the global stacking context.

Should look like this, post-fix.

Screen Shot 2021-11-10 at 11 42 14 AM

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/static/sass/custom/combo-box.scss
+++ b/crt_portal/static/sass/custom/combo-box.scss
@@ -1,3 +1,9 @@
+.usa-combo-box {
+  // Create a new stacking context to hold generated combo box elements
+  // that have its own z-index
+  z-index: 1;
+}
+
 .usa-combo-box input {
   border-radius: 3px;
 }


billyhunt (approved):
Works as intended on both the dashboard and the view pages.

louh:
Just leaving a note for myself - I looked into whether this change is worth patching upstream on USWDS, and it actually introduces another problem if there is another combo-box (or generically speaking, another stacking context) below this one: making the button UI not appear other elements can then cause the dropdown to not properly appear over other UI. We don't have that problem here, but it could also mean that (a) we would need to find a follow-up solution in case it does become a problem, or (b) we still need to see if a different solution could be found in a future update to USWDS. I'll open that conversation there too.

billyhunt:
Good work!

louh:
@billyhunt I'm going to close this in favor of #1038, where I have to make a few changes to the "assigned to" UI anyway (https://github.com/usdoj-crt/crt-portal-management/issues/1140), and I'm going to address this in the modal z-index rather than the combo box z-index.

PR #1047, opened 2021-11-22
Bh/1087 add materialize view

billyhunt:
[Link to ZenHub issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1145)

## What does this change?

When looking at the sql queries for the /form/view page, one sticks out with large datasets.  The email_report_count.  

image (2)

Notice the two red queries related to email_report_count.  The total time of just those two queries is almost 2.5 seconds.  Thats a lot of time for a user to wait for the page to load.  

email_report_count is a view, meaning it is computed every time the page loads.  Nothing is cached.  We decided to change this view to something called a "materialized view", which caches the data.  Look at the same page query performance when we use a materialized view.  

image (3)

About .5 seconds.  Still not great, but MUCH better.  2 seconds of page load time shaved off.  

This query will also save time on the form/view/{index} page for individual pages.

The downside of using a "Materialized View", is we need to recompute to make it accurate. We will update every time someone creates a report, or updates an email in the existing report.  This will probably slow down the report creation process, but we feel that it is worth the cost to save time displaying report information because

1. Reading is done more often than writing
2. The amount of time saved on the report read is greater that the time spent writing.

## Rollback plan:
If for some reason we need to roll back from this database migration, I have created a PR to quickly do it.

If a rollback is needed, merge the following PR in.  https://github.com/usdoj-crt/crt-portal/pull/1048

This will remove the materialized view and replace it with the regular email_report_count view.  

### Author

### Reviewer

## Notes for reviewer:


louh:
This would be a good candidate for a squash and merge from the GitHub interface IMO.

billyhunt:
I am defaulting to squash and merge now.

PR #1053, opened 2021-11-24
Update to USWDS 2.10.3

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1146)

## What does this change?

Update to US Web Design System 2.10.3, the next incremental update in an ongoing effort to bring our design system to the latest and greatest components and enhancements. The full official changelogs are here: [2.10.0](https://github.com/uswds/uswds/releases/tag/v2.10.0), [2.10.1](https://github.com/uswds/uswds/releases/tag/v2.10.1), [2.10.2](https://github.com/uswds/uswds/releases/tag/v2.10.2) (Note that [2.10.3](https://github.com/uswds/uswds/releases/tag/v2.10.3) is a bug-fix release, no changes were introduced.)

The following changes are relevant to us:

- Banner content no longer flashes "open" on page load.
- Checkbox focus borders now display at the correct size on Firefox. This fixes a previously known issue, https://github.com/usdoj-crt/crt-portal-management/issues/1074.
- The combobox component should now have an accessible name.
- Behind the scenes, USWDS is introducing an SVG icon build system and a new directory, `usa-icons`, to contain them. We may slowly adopt this over time. SVG icon build systems can improve the front-end loading experience.
- Updates to `uswds-gulp` and Node.js dependencies.

n/a

### Author

### Reviewer

## Notes for reviewer:


crt_portal/static/sass/custom/landing.scss
```diff
@@ -154,10 +154,11 @@ $arrow-height: 24px;
...
         }
 
+        // Override USWDS styles
         a {
-          color: #162e51;
-          padding-top: 0px;
-          padding-left: 0.5rem;
+          color: #162e51 !important;
+          padding-top: 0 !important;
+          padding-left: 0.5rem !important;
```
billyhunt, line 161: Could this be done by narrowing the selector?

crt_portal/static/sass/custom/landing.scss
```diff
@@ -181,7 +182,7 @@ $arrow-height: 24px;
   .crt-landing--menu_link {
     color: #162e51 !important;
     font-weight: normal !important;
-    padding: 1rem 0.25rem;
+    padding: 1rem 0.25rem !important;
```
billyhunt, line 185: See above.

billyhunt (changes requested):

crt_portal/static/sass/custom/landing.scss
louh: Not really - at least, not in a way that's even worse-looking than `!important` - although it's worth looking at why the thing I'm overriding needs to exist. I might need to redo part of the navigation and unwind some of the complexity that's already there. I'll take a closer look.

billyhunt (approved):
After talking with @louh, I agree his !important solution is the best for now.  We might want to circle back and consider refactoring some solutions down the line.

louh:
I'll leave some notes here for future reference in case we need it.

So here's what's going on: the `!important` is there to override this very-specific CSS selector: `.usa-nav__primary a:not(.usa-button):not(.crt-button):not(.outline-button):not(.filter-tag)`

The stacked `:not` pseudo-selectors are a result of our Sass extends of the `.usa-button` class. They are unexpectedly piling onto the original `.usa-nav__primary a:not(.usa-button)` selector in the USWDS.

We had three options:

- Use an `id`  in the DOM as a selector - which trades off one problematic CSS tactic with another
- Rewrite button styles to not use Sass extends - which is a much more significant lift
- Keep the `!important` in place. Right now, this seems to be the best solution, as it constrains to the "code smell" to just the parts where it needs it.

PR #1057, opened 2021-12-02
fix(intake): remove extraneous screen reader tag

louh:
## What does this change?

From screen reader testing with Terri today: when the intake format tooltip is being read correctly (in Jaws 18, but there seems to be other issues with Jaws 2020 and NVDA, we will need to follow up with that separately) it will read the intake format twice. This is because of an extraneous `.sr-only` tag. We can remove that and let the USWDS tooltip component handle the screen reader text.

n/a

### Author

### Reviewer

## Notes for reviewer:


ilodidoj (approved):
Checked locally. LGTM. @louh ready for your merge.

PR #1078, opened 2022-01-10
Update Django 3.2

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1171)

## What does this change?

As part of a larger process to allow grouping of AND/OR search term combinations in intake filters (https://github.com/usdoj-crt/crt-portal-management/issues/1165), this PR updates Django to the current long-term support (LTS) version, v3.2. The allows us to use Postgres v11+ built-in "websearch" functionality at the database level, once Postgres is updated (next step: https://github.com/usdoj-crt/crt-portal-management/issues/1172).

Note that Django would have needed to be upgraded anyway, as our previous version, v2.2, reaches [end of the support cycle in April 2022](https://www.djangoproject.com/download/#supported-versions). Updating Django allows us to stay on top of new security updates through 2024.

The following is a run-down of all changes made to update our project from Django v2.2 to v3.2. Changes were made only to programming logic, there have been no user-facing or product changes in the application. All existing functionality is expected to behave the same as before.

- The `postgresql_psycopg2` database backend was a deprecated alias for `postgresql`, so it has been updated to point at the actual package.
- The `USE_TZ` setting can be re-enabled as it no longer conflicts with a previous Python update.
- Compressor configuration syntax has changed, however, we do not actually use this right now. I've left it in place with updated settings in case it is re-enabled in the future.
- Django now auto-detects default configuration, and the `default_app_config` setting was deprecated.
- Automatic primary keys in Django 3.2+ uses `django.db.models.BigAutoField`. To future-proof the default and preserve original behavior, existing models were set to use `default_auto_field = 'django.db.models.AutoField'`. Also, if I didn't do that, the app can get really noisy with warnings about it. For the Actstream package, this meant I had to subclass the plugin to provide that value, and this shouldn't affect any other functionality.
- Minor: a `BoundField` import has changed.
- The `ifequal` tag in templates has been removed. Using the `==` operator will do the same thing.
- Template paths in `ChoiceWidget` components can no longer use `../` relative paths. They've been updated so they can be properly appended to Django's template paths.
- In tests, `assertEquals` is an alias of the non-plural `assertEqual`, but it was deprecated.
- A new `ModelChoiceIteratorValue` was introduced as part of making drop-down select options, and it behaved differently than it has in the past. I made a fix to tweak the new behavior back to the way we expected to use it, but included a note because Django will be updated in v4 to address our particular use case.
- HttpResponse objects now have a `headers` property. While the old way (setting headers directly on the response object) still works, I think using the `headers` property provides more clarity about what we are doing.

**For local development:** Your existing local instance can be upgraded in place by running `docker-compose up --build`. If there are any issues with running the project after this, try removing old images (e.g. `docker system prune`) and building fresh images. You should not need to drop or rebuild your local volumes (including databases).

n/a

### Author

### Reviewer

## Notes for reviewer:


ilodidoj:
@louh I am reviewing this PR now on my local as well. Let you know by noon. CC: @billyhunt

billyhunt:
+1

billyhunt (approved):

PR #1101, opened 2022-02-08
Advanced search functionality (aka "search v2.0")

louh:
[Link to issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1173)

## What does this change?

Updates the "personal description" search filter to allow new search operators provided by PostgreSQL's "websearch" functionality, with caveats.

Here's what we can do in the search box:

- **Existing behavior:**
  - `AND` (default behavior). Searches for `mask vaccine` or `mask AND vaccine` return results containing both words.
  - `OR`. Searches for `mask OR vaccine` return results containing either word.
- **New behavior:**
  - Negation with `-` in front of a word. Searches for `mask -vaccine` return results containing "mask" but not "vaccine."
  - Adjacent word search with quotation marks. Searches for `"mask vaccine"` return results where the words must be in the same order. (See also caveats.)
  - Grouping with parentheses. Searches for `mask AND (vaccine OR shot)` allow changing the order of operations for search operators. This returns results containing both "mask" and "vaccine", along with results containing both "mask" and "shot". (See also caveats.)

### Caveats (for end users)

**Adjacent word search ignores "stop" words.** (Stop words are common English words such as "a" or "the", which include prepositions and articles.) I originally called this "exact phrase search" but that's not exactly correct because that would imply a search for "royale with cheese" must include those same three words in that order. In actuality, PostgreSQL will also return a result containing "royale and cheese" as valid. This behavior can create unexpected or useful results, depending on the use case.

**Grouping cannot be nested.** It's conceivable that someone might want to construct a query such as `mask AND (vaccine OR (shot AND pfizer))` but the query will be flatted into one set of parentheses. This is a limitation of the PostgreSQL search functionality.

**Adjacent word search and grouping cannot be used in the same query.** Someone might want to construct a query such as `mask AND (vaccine OR "covid shot")` but this will not work. ~~The search results drop the parentheses and will prioritize the adjacent word search.~~ (EDIT: Search results may be unexpected, it's not always obvious what will take priority.) This is also a limitation of PostgreSQL search functionality, where a workaround is theoretically possible but outside of the scope of this ticket.

### Technical notes

PostgreSQL's "websearch" functionality actually does not allow the grouping of search terms with parentheses, something that I did not catch until after we were able to put everything together after the Django and PostgreSQL updates. This misunderstanding occurred because the [Django documentation on SearchQuery](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/search/#searchquery) implied in their examples that the parentheses would be meaningful operators in websearch, however, [PostgreSQL only supports grouping in raw search queries, not websearch](https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES).

The current workaround to achieve our desired result is to do a simple string replacement on incoming search queries if a set of parentheses is detected. This converts incoming web search syntax into raw search syntax so that the results contain the grouping we want.

Because of this string replacement technique, I don't have a way to also handle adjacent word search (with quoted text). There are ways of doing this, such as parsing incoming queries into an abstract syntax tree and re-assembling a raw text search query, but I decided that given our original scope (to support grouping of queries) that it would not be feasible to pursue that strategy during this sprint.

Websearch is still useful for us, as other forms of text search (such as adjacent words) will be automatically parsed, and its ability to gracefully handle malformed search syntax. Raw search is very picky about syntax and won't hesitate to throw errors if the syntax isn't well-formed.

This means, in cases where we provide a raw search query that cannot be parsed, we continue to fall back on websearch functionality. This allows us to catch syntax errors. Even if the search results may not be what a user expects, it'll be better than throwing up an error page, and gives the user an opportunity to refine their query to something that might return better results.

### Deprecations

The undocumented workaround for creating an "OR" query by adding additional `?violation_summary=` parameters in a URL string has been removed, in favor of using the "OR" operator in the search query. Users who were depending on this previous functionality should be encouraged to revise bookmarks or links so that they are using the currently documented search functionality.

n/a

## Checklist:

I would prefer if people try testing search queries on their own to try and break this, instead of running through a sequence of tests I know will work. But to get started, here's what I would try:

- [ ] Using the pro-form, create a series of reports with short personal descriptions:
    - "mask"
    - "mask vaccine"
    - "mask shot"
    - "mask vaccine shot"
    - "vaccine shot"
- [ ] Test query `mask AND (vaccine OR shot)` => 3 results.
- [ ] Test query `mask -shot` => 2 results.
- [ ] Test query `"vaccine shot"` => 2 results.
- [ ] Test query `mask -"vaccine shot"` => 3 results.
- [ ] Test query `mask -(vaccine OR shot)` => 1 result.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/filters.py
+++ b/crt_portal/cts_forms/filters.py
@@ -1,5 +1,6 @@
 # Class to handle filtering Reports by supplied query params,
 # provided they are valid filterable model properties.
+import re
 import urllib.parse
 from datetime import datetime
 
@@ -172,6 +173,10 @@ def _make_search_query(search_text):
         search_text = search_text.replace(' AND ', ' & ')
         search_text = search_text.replace(' OR ', ' | ')
         search_text = search_text.replace(' -', ' !')
+        # In between search tokens that don't have operators, insert &
+        # e.g. "foo -(bar | baz qux)" => "foo & !(bar | baz & qux)"
+        search_text = re.sub('([a-zA-Z)"]) ([a-zA-Z("!])', r'\1 & \2', search_text)
+        # Note: the only search type we don't handle here are exact phrase quotes.
         with connection.cursor() as cursor:
             try:
                 # This can still create syntax errors, so we ask the db to validate the
--- a/crt_portal/cts_forms/filters.py
+++ b/crt_portal/cts_forms/filters.py
@@ -4,6 +4,7 @@ import urllib.parse
 from datetime import datetime
 
 from django.contrib.postgres.search import SearchQuery
+from django.db import connection
 
 from .models import Report, User
 from actstream import registry
@@ -165,10 +166,23 @@ def dashboard_filter(querydict):
 
 
 def _make_search_query(search_text):
+    # Websearch will drop parentheses from query. So if a set of parentheses is
+    # detected, we attempt to convert it to a tsquery
     if '(' in search_text and ')' in search_text:
         search_text = search_text.replace(' AND ', ' & ')
         search_text = search_text.replace(' OR ', ' | ')
-        query = SearchQuery(search_text, config='english', search_type='raw')
+        search_text = search_text.replace(' -', ' !')
+        with connection.cursor() as cursor:
+            try:
+                # This can still create syntax errors, so we ask the db to validate the
+                # query for us. This query only parses the input, it doesn't incur the
+                # performance hit of executing the search twice. If the query isn't
+                # valid, we catch the error and execute a websearch instead, which is
+                # more forgiving with syntax errors.
+                cursor.execute("SELECT to_tsquery('english', %s);", [search_text])
+                query = SearchQuery(search_text, config='english', search_type='raw')
+            except Exception:  # catch *all* exceptions
+                query = SearchQuery(search_text, config='english', search_type='websearch')
     else:
         query = SearchQuery(search_text, config='english', search_type='websearch')
 
--- a/crt_portal/cts_forms/filters.py
+++ b/crt_portal/cts_forms/filters.py
@@ -128,10 +128,7 @@ def report_filter(querydict):
             elif filter_options[field] == '__gte':
                 kwargs[field] = querydict.getlist(field)
             elif filter_options[field] == 'violation_summary':
-                combined_or_list = []
-                for vs_filter in filter_list:
-                    combined_or_list += vs_filter.split(" OR ")
-                combined_or_search = _combine_term_searches_with_or(combined_or_list)
+                combined_or_search = _combine_term_searches_with_or(filter_list)
                 qs = qs.filter(violation_summary_search_vector=combined_or_search)
     qs = qs.filter(**kwargs)
     return qs, filters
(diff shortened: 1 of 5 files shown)


billyhunt:
@louh , first of all great effort on bringing this search to the next level.  I think its awesome.  I do have some concerns about some of the limitations, or at least how we are handling the limitations.  Some of the complicated searches we could do will return unexpected results.  For example this one.  
image
It returns the one result it should exclude.  Its definitely an edge case, but I hate incorrect results returned (as opposed to saying "don't do this" in the UI or having it work.  

On the other hand, not many people are doing these advanced searches, so I don't want to put too much time into it.  But should we have a message to users saying like "we cannot complete your search" when they do something strange we don't support?

What do you think @ilodidoj?

louh:
Some things we've found in testing the above query, `(shot AND mask) -"describe visit"`:

- I originally assumed that parentheses would be dropped and the quoted text would be prioritized instead, but that's not what happens in all cases, including here. I edited the original post to no longer make any claims about what the system might do when it encounters a query that combines parentheses and quotation marks. The only guaranteed claim I can make is "results might be unexpected."
- Sometimes advanced queries can be written in a simpler form to return expected results. The given example here can be written as `shot mask -"describe visit"`, with minimal operators, as there's no need for grouping to indicate order of operations here. Simplifying search queries guarantees correct results.

Based on @billyhunt's recommendation, I am trying this UI change which catches this sort of query during input, providing a warning and blocking the user from performing the search:

Screen Shot 2022-02-11 at 9 40 15 AM

Thoughts? @ilodidoj @blaughman @randyabramson

randyabramson:
Thanks @louh - for my 2 cents, the most important thing we can provide here is the grouping of terms. This sort of query represents what we're after:

==
veteran and (car OR towed OR lease)
==

The info above you have about exact word matching is important, but we can explain the limitations of the system in the 'help modal'.

Is the query I outlined above achievable?

billyhunt (approved):
Looks good to me.  Great work!

chinelo-ikejimba:
@louh @billyhunt let's put show the error message in an alert for design consistency
[Design Handoff Link](https://www.figma.com/file/g9rb8IUGcgiJNJUNLrwG1T/CRT-Aug-2021-Working-pages-(Copy)?node-id=625%3A6935)


Screen Shot 2022-02-11 at 2 07 56 PM

louh:
@randyabramson Yes, the example query you provided can be done. The ability to do group terms like `a AND (b OR c [...OR x...])` is the primary goal.

I just added another test suite to make sure we results are expected for more than 2 OR queries.

We might be too late down this path to suggest this, but we could also avoid mentioning we can do quoted searches or negated terms, even though it came with PostgreSQL.

louh:
@chinelo-ikejimba That's a good recommendation, thank you! Using the slim alert component will cut down on some custom code that I made up.

The slim alert we have currently uses italicized text, like so:

Screen Shot 2022-02-11 at 3 19 32 PM

Any objection to leaving it italicized here?

Screen Shot 2022-02-11 at 3 19 39 PM

chinelo-ikejimba:
@louh italicizing is fine. I'll update the bulk actions error message to follow the same style. Great job!

louh:
Thanks @chinelo-ikejimba!

@billyhunt, I know you approved already but since we've made some changes, mind taking another glance? cc @ilodidoj

billyhunt:
Tested.  Looks good!

PR #1131, opened 2022-03-10
change settings to allow js and css caching

billyhunt:

## What does this change?

Users are running into problems when code is updated, that their browsers are not reset.  We need to have a way to "cache bust" this by changing the file names of changed assets, so the browser will reload these assets.  

The simplest way to do this is from within django.  Since we are using s3, we need to use S3ManifestStaticStorage as referenced [here](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html)

## Checklist:

This cannot be looked at locally.  The only way to test this is to push to dev, and see if the hashes are created.  

### Author

### Reviewer

## Notes for reviewer:


ilodidoj:
@billyhunt there may be some more additional changes needed.  Please research the code base for s3boto3.  It might have impact on how to access the AWS bucket. All buckets use the same authentication class to access the bucket.

ilodidoj (approved):
let's merge to dev to validate AWS bucket files for this changes.

PR #1200, opened 2022-06-30
Update db migration to show correct values

billyhunt:
[Link to ZenHub issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1309)

## What does this change?

When rebuilding the database, one of the Protected Class fields (Now called Personal characteristics) is missing because of a bug in a migration.  This PR will add it the field back if it is missing, so if we ever had to rebuild the database on dev, staging, or production, it would be there.  

Note this will only change the field if it is incorrectly assigned to this specific value.  Anything else will be ignored, so there will be no effect on current dev, staging, or prod data, nor will there ever be, unless we are forced to rebuild those systems. If we do have to rebuild, then the form data will be built correctly.

Before: (No Disability Option)
image

After: (with Disability Option)
image

## Checklist:

+ [x] To test locally, run this code, and make sure "Disability (including temporary or recovered and including HIV and drug addiction)" is in both the report in the Personal Characteristics section, and the Pro Form.  

### Author

### Reviewer

## Notes for reviewer:


JackRyan1989 (approved):
Tested locally, looks good to me!

PR #1219, opened 2022-08-01
Pagination Options on Search

eileen-nava:
Issue: [Enable users to customize how many reports display on /form/view](https://github.com/usdoj-crt/crt-portal-management/issues/1281#)

I am creating a draft PR in order to transfer my work to @billyhunt.  This work is mostly done, however, it still needs styling in order to match the associated design.

Update to work.  To test, 

+ [ ] Click the Kabob menu above the right side of the table at /form/view
image

+ [ ] Try selecting different page sizes, and check the number of records that are returned. 

image


JackRyan1989 (changes requested):
@billyhunt I'm seeing a slight display issue in Firefox only. 
Screen Shot 2022-08-05 at 4 38 04 PM

You can see that a little bit of the text in the first option element is appearing. I think you can fix this by increasing the padding to 1.2rem in this rule: `.usa-combo-box__input, .usa-input, .usa-input-group, .usa-range, .usa-select, .usa-textarea`

Looks like that rule is set in _border.scss, and it will affect the Personal Description input box as well. So maybe make a new rule for just this box?

billyhunt:
> @billyhunt I'm seeing a slight display issue in Firefox only. Screen Shot 2022-08-05 at 4 38 04 PM
> 
> You can see that a little bit of the text in the first option element is appearing. I think you can fix this by increasing the padding to 1.2rem in this rule: `.usa-combo-box__input, .usa-input, .usa-input-group, .usa-range, .usa-select, .usa-textarea`
> 
> Looks like that rule is set in _border.scss, and it will affect the Personal Description input box as well. So maybe make a new rule for just this box?

I am not seeing this.  Can you let me know the browser and window size?

billyhunt:
I don't think we need to worry about firefox here, as this is an internal tool and is only used on edge.

ilodidoj (changes requested):
@billyhunt CI/CD showing failed tests. crt_portal/cts_forms/tests/test_views.py", line 395, in test_total_pages
    self.assertTrue(f'{num_records} of {num_records} records' in self.content). Please check this test on local docker test before pushing or merging to dev.

JackRyan1989:
> I don't think we need to worry about firefox here, as this is an internal tool and is only used on edge.

Ok cool. It looks ok on everything except Firefox.

billyhunt:
@ilodidoj and @JackRyan1989 Do we feel like this is ready to go?  If so can you mark it as such?

JackRyan1989 (approved):
I'm ok with approving this

ilodidoj:
@billyhunt @JackRyan1989 doing local testing, update this in a few min.

ilodidoj:
@billyhunt according to design handoff should there be option to view 100 records as well? If so, please add that to per_page html.  CC: @JackRyan1989

billyhunt:
@ilodidoj We decided against that to start, as it could really bog the system down.

ilodidoj (approved):

ilodidoj (approved):
LGTM

PR #1261, opened 2022-11-04
Add export to csv action to TMS section of admin page.

billyhunt:
[Link to ZenHub issue.](https://github.com/usdoj-crt/crt-portal-management/issues/1417)

## What does this change?

As a user, we would like to be able to export TMS email data to track whats going on.  This PR adds an admin action to export the data to CSV.

image

Exported CSV File
image
## Checklist:

+ [x] To test locally, you need TMS data.  To get that, run ``` docker-compose run web python crt_portal/manage.py create_mock_email_info ``` two or three times to create two or three records. 
+ [x] Go to the tms section of the admin panel.  Check the rows, click the actions dropdown, and download them by clicking "go"
+ [x] Check the downloaded report.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/tms/admin.py
+++ b/crt_portal/tms/admin.py
@@ -7,7 +7,7 @@ from django.http import StreamingHttpResponse
 
 from .models import TMSEmail
 
-TMS_FIELDS = ['id', 'tms_id', 'recipient', 'subject', 'body', 'recipient', 'status', 'error_message', 'report_id', 'created_at', 'completed_at', 'purpose']
+TMS_FIELDS = ['id', 'tms_id', 'subject', 'body', 'recipient', 'status', 'error_message', 'report_id', 'created_at', 'completed_at', 'purpose']
 
 logger = logging.getLogger(__name__)
 
--- a/crt_portal/tms/admin.py
+++ b/crt_portal/tms/admin.py
@@ -1,9 +1,38 @@
-from cts_forms.admin import ReadOnlyModelAdmin
+import logging
+from cts_forms.admin import ReadOnlyModelAdmin, csv, Echo, iter_queryset, format_export_message
 from django.contrib import admin
 from django.utils.html import format_html
 from django.urls import reverse
+from django.http import StreamingHttpResponse
 
 from .models import TMSEmail
+
+TMS_FIELDS = ['tms_id', 'recipient', 'report', 'status', 'created_at', 'completed_at', 'purpose']
+
+logger = logging.getLogger(__name__)
+
+
+def _serialize_tms(data):
+    """Preserve headers while rendering TMS_FIELDS for inbound actions"""
+    if isinstance(data, TMSEmail):
+        return [getattr(data, field) for field in TMS_FIELDS]
+    return data
+
+
+def export_tms_as_csv(modeladmin, request, queryset):
+    """
+    Stream actions as csv
+    Log all use
+    """
+    writer = csv.writer(Echo(), quoting=csv.QUOTE_ALL)
+    iterator = iter_queryset(queryset, TMS_FIELDS)
+    response = StreamingHttpResponse((writer.writerow(_serialize_tms(tms)) for tms in iterator),
+                                     content_type="text/csv")
+    response['Content-Disposition'] = 'attachment; filename="tms_export.csv"'
+
+    logger.info(format_export_message(request, queryset.count(), 'tms log entries'))
+    return response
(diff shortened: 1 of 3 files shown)


ilodidoj (commented):
@billyhunt Good work on this.  Please note, this will be similar to report export.  
Query view list will remain as is, but the actual csv export will have more fields from the TMS detail page.  Export to CSV Fields should include TMS id, Report id, Subject, Body, Recipient, Created at, Completed at, Status, Purpose, Error message.  That is all the fields available in the TMS record detail view page.

billyhunt:
Ok!  I will update with the rest of the fields.

billyhunt:
All the fields are added, as the generator is updated to reflect the dates more like those in the reports section.  Note it will still be a little different until we run it in dev and see how it works.  (for example, the subject and body are auto generated, not real emails)

crt_portal/tms/admin.py
```diff
@@ -1,10 +1,39 @@
...
+TMS_FIELDS = ['id', 'tms_id', 'subject', 'body', 'recipient', 'status', 'error_message', 'report_id', 'created_at', 'completed_at', 'purpose']
+
+logger = logging.getLogger(__name__)
+
+
+def _serialize_tms(data):
+    """Preserve headers while rendering TMS_FIELDS for inbound actions"""
+    if isinstance(data, TMSEmail):
+        return [getattr(data, field) for field in TMS_FIELDS]
+    return data
```
JackRyan1989, line 19: This feels like it could turn into a bug, maybe? If something is not an instance of the TMSEmail model, I feel like we should return an exception, right? We're trying to serialize a very specific thing, and I think if the data coming in isn't that thing, then we shouldn't serialize it? Happy to chat more.

crt_portal/tms/admin.py
```diff
@@ -1,10 +1,39 @@
...
+
+
+def export_tms_as_csv(modeladmin, request, queryset):
+    """
+    Stream actions as csv
+    Log all use
+    """
+    writer = csv.writer(Echo(), quoting=csv.QUOTE_ALL)
+    iterator = iter_queryset(queryset, TMS_FIELDS)
+    response = StreamingHttpResponse((writer.writerow(_serialize_tms(tms)) for tms in iterator),
```
JackRyan1989, line 29: Since the iteration is dependent on the TMS fields, if we pass some other (non-TMS Model) array in there and it gets returned by serialize_tms, isn't this for loop gonna break?

crt_portal/tms/admin.py
```diff
@@ -17,6 +46,7 @@ def tms_detail_url(self, obj):
         return format_html('<a href="%s">View & update</a>' % (reverse('tms:tms-admin-message', kwargs={"tms_id": obj.tms_id})))
```
JackRyan1989, line 46: Is the view for the TMS email not done yet? Here's what I see when I click on this link:

Screen Shot 2022-11-04 at 3 41 31 PM

JackRyan1989 (changes requested):

crt_portal/tms/admin.py
billyhunt: This export function is specifically for tms, so I don't think we would use for other models.  Its basically a copy of "export_actions_as_csv" over in tms_forms/admin.py

crt_portal/tms/admin.py
billyhunt: I am copying the pattern over in cts_forms/admin.  My guess is they coded it that way so if you added new fields to the model, it would not break if python had trouble reading those fields, like if it was something complicated.  (see how they are doing it in reports)

This _serialize function is only for tms models though.  Do you think we should extend it to be more general.  The pattern so far is to have a serialize function and export function that is specific for each model.  I personally think that is the right way to go for clarity (like report is different than actions).  

What do you think?

crt_portal/tms/admin.py
billyhunt: This is a known bug, and outside the scope of this PR.

crt_portal/tms/admin.py
JackRyan1989: Sounds good! This answers my question. Thanks @billyhunt

crt_portal/tms/admin.py
JackRyan1989: Gotcha! Thanks @billyhunt

JackRyan1989 (approved):
Looks good to me! Concerns resolved ion one on one conversation.

PR #1302, opened 2023-01-18
Fix CircleCI Failures after Python 3.10.8 Upgrade

okkays:
## What does this change?

Fixes our CircleCI failures following the update to python 3.10.8.

As a few versions got bumped during the switch (namely node and the browsers container image), there were a few things going on here that contributed to the failures:

### New path for `google-chrome`

- 🌎 Pa11y and e2e tests use Puppeteer to run chrome
- ⛔ Those tests were failing on circleci because puppeteer
  couldn't find the Chrome binary in the new python3.10.8-browsers image.
- ✅ We now explicitly say where to find the Chrome binary when running on
  CircleCI (`/usr/bin/google-chrome`)

### CURL not retrying

- 🌎 In CircleCI, before running a11y tests, we need to wait for the dev server to come up. We currently do that using `curl`
- ⛔ Curl's `retry-connrefused` flag isn't actually retrying on CCI - it tries
  once and fails
- ✅ We now retry _all_ errors, instead, giving the dev server time to come up before failing the tests

### 127.0.0.1 vs localhost

- 🌎 We were using `127.0.0.1` explicitly instead of `localhost`.
- ⛔ `127.0.0.1` only works for ipv4. Some versions of node, and thus pupeteer,
  now support ipv6 by default (see:
  https://github.com/puppeteer/puppeteer/issues/8873). This is not the full solution, but ipv6 is the way of the future, and we should allow for it now that Node prefers it.
- ✅ We now use localhost instead, which can resolve to either
  ipv4 or ipv6 as needed.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,13 +119,13 @@ jobs:
       - run:
           name: Run pa11y-ci accessibility check
           command: |
-            curl --retry-delay 5 --retry 10 --retry-connrefused http://127.0.0.1:8000
+            curl --retry-delay 5 --retry 10 --retry-connrefused http://localhost:8000
             npm run test:a11y-without-report
       - store_artifacts:
           path: ./pa11y-screenshots
       - run:
           name: E2E tests
-          command: pipenv run pytest crt_portal/cts_forms/tests/integration/*.py --base-url=http://127.0.0.1:8000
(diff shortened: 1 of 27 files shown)


JackRyan1989 (approved):
This looks good to me!

PR #1317, opened 2023-02-02
Add flag to hide response templates from dropdown and automate hiding for deleted md file templates

kewitham:
[Issue #1460](https://github.com/usdoj-crt/crt-portal-management/issues/1460)

## What does this change?
This PR adds a field to the ResponseTemplate model with a flag to hide the template in the portal response template select dropdown and a field to indicate which templates are user created (added in the admin console or migrated).
It also adds logic to hide templates represented by an md file once that file is deleted.

**Admin Page**
image

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/management/commands/update_response_templates.py
```diff
@@ -47,14 +54,26 @@ def handle(self, *args, **options):
...
 
                     letter, created = ResponseTemplate.objects.update_or_create(title=letter_id, defaults=letter_data)
 
                     if created:
                         self.stdout.write(self.style.SUCCESS(f'Created response template: {letter.title}'))
                     else:
                         self.stdout.write(self.style.SUCCESS(f'Updated response template: {letter.title}'))
+
+        objects_without_templates = filter(self.template_exists, ResponseTemplate.objects.all())
+        for object in objects_without_templates:
```
okkays, line 72: I _think_ you can just iterate over the query:

```py
for object in ResponseTemplate.objects.filter(...):
```

Rather than using `.all` - the difference being, if you use `all`, Django can't do any memory efficiency tricks or short circuiting and must fetch everything all at once.

[Here's a bit of a dated, but still interesting, deep dive into the pros and cons](https://blog.etianen.com/blog/2013/06/08/django-querysets/)

crt_portal/cts_forms/management/commands/update_response_templates.py
```diff
@@ -47,14 +54,26 @@ def handle(self, *args, **options):
...
+            if not object.is_user_created:
```
okkays, line 73: You can also add this to the filter above to be done at the sql, rather than python, level

crt_portal/cts_forms/migrations/0161_auto_20230202_1257.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.2.16 on 2023-02-02 17:57
```
okkays, line 1: This is going to be annoying, but can you rename this file?

To name when you migrate:
```
docker-compose run web python /code/crt_portal/manage.py makemigrations -n my_migration_name
```

crt_portal/cts_forms/migrations/0161_auto_20230202_1257.py
```diff
@@ -0,0 +1,29 @@
...
+        migrations.AlterField(
+            model_name='report',
+            name='crt_reciept_year',
+            field=models.PositiveIntegerField(blank=True, null=True, verbose_name=django.core.validators.MaxValueValidator(2023)),
+        ),
+        migrations.AlterField(
+            model_name='report',
+            name='last_incident_year',
+            field=models.PositiveIntegerField(blank=True, null=True, verbose_name=django.core.validators.MaxValueValidator(2023, message='Date can not be in the future.')),
+        ),
```
okkays, line 28: Ohhhhh I see why it's generating these - because it's a new year:
```
MaxValueValidator(2023, message='Date can not be in the future.')
```

crt_portal/cts_forms/migrations/0162_responsetemplate_show_in_dropdown.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.16 on 2023-02-02 20:57
```
okkays, line 1: Can you bundle this with the migration above? It helps to keep a cleaner record of what was migrated with what PR

crt_portal/cts_forms/tests/test_api.py
```diff
@@ -180,14 +180,14 @@ def test_response_detail_url(self):
         response = self.client.get(self.url)
         self.assertEqual(response.status_code, 200)
         self.assertTrue("false" in str(response.content))
-        self.assertFalse("true" in str(response.content))
+        self.assertTrue("/api/responses" in str(response.content))
```
okkays, line 183: I'm having trouble understanding why this changed - can you help explain?

okkays (changes requested):

crt_portal/cts_forms/tests/test_api.py
kewitham: The response content used to not have the value "true" in it but now it sometimes does because of the new is_user_created field (which is often set to true) so I added a different content check based off of the checks performed elsewhere in this file.

crt_portal/cts_forms/management/commands/update_response_templates.py
```diff
@@ -10,6 +10,13 @@ class Command(BaseCommand):  # pragma: no cover
     help = 'Adds new response templates or updates existing ones'
 
     templates_dir = os.path.join(settings.BASE_DIR, 'cts_forms', 'response_templates')
+    template_ids = []
+
+    def template_exists(self, object):
+        if object.title in self.template_ids:
+            return
+        else:
+            return object
```
okkays, line 19: I think this can be removed now - looks like it's not used since the refactor below?

```suggestion
```

crt_portal/cts_forms/tests/test_api.py
okkays: ah okay, thanks!

in that case, I'm not quite sure how this tests the response detail url - can you assert that something from SAMPLE_REPORT_1 is present in this response?

okkays, line 183: Also, something I've seen throughout these files, that we can change as we see them - `assertContains` is a thing.

So, we can replace:

```
self.assertTrue(member in container)
```

with

```
self.assertContains(container, member)
```

The difference being that the latter will give a much better error message ("member" was not present in "container" - with their actual values) versus "false is not true"

okkays (changes requested):

okkays (approved):
LGTM!

Can you comment with a list of QA steps for when this is merged?

e.g.:

- [ ] Go to /admin/responsetemplates
- [ ] Confirm that a, b, and c are hidden
- [ ] Confirm that x, y, and z are NOT hidden
- [ ] Create a new response template, and confirm it's not hidden

kewitham:
Post-merge QA steps:
1. Navigate to a response detail page and click to "Contact complainant" to verify that the DRS templates are not visible in the dropdown.
2. Verify that the "Trending - General COVID Inquiries" template is in the dropdown.
3. Navigate to '/admin/cts_forms/responsetemplate/' and verify that all DRS templates are still in the console but have the "Show in select template dropdown" unchecked.

cc: @okkays @ilodidoj

PR #1321, opened 2023-02-06
Add constant writer icon to view all table

kewitham:

## What does this change?
This PR adds a constant writer icon to the view all reports table to flag reports from users whose contact email addresses had already received a repeat writer email.
image

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/views.py
```diff
@@ -182,6 +183,38 @@ def _related_reports_count(report):
...
+    with connection.cursor() as cursor:
+        try:
+            cursor.execute(
+                """  select act.description as description, r.contact_email as email
+                from cts_forms_report r
+                left join actstream_action act on concat('Email sent: ''CRT - Constant Writer'' to ', r.contact_email, ' via govDelivery TMS') = act.description
+                WHERE act.verb = 'Contacted complainant:'
+                AND r.contact_email IN %s
+                order by act.timestamp desc
+                """, params=[report_emails])
```
okkays, line 209: FYI @ilodidoj 

Katie and I paired on this query and proved to ourselves this isn't possible with the ORM. Even if we were to construct a view, we'd need to use raw SQL to construct the view, and the data would be 24 hours delayed (where here the constant writer icon will appear instantly with this code)

- To get Django to construct a join, you need a ForeignKey between the two join targets.
- The ForeignKey must be based off of an existing field on the table.
- This join's "foreign key" is a constructed field (built with CONCAT)
- Django doesn't allow constructed fields like this to be defined at the model level
- So, we can't define a ForeignKey based on a constructed field

crt_portal/cts_forms/views.py
```diff
@@ -182,6 +183,38 @@ def _related_reports_count(report):
...
+        return
+    with connection.cursor() as cursor:
+        try:
+            cursor.execute(
+                """  select act.description as description, r.contact_email as email
+                from cts_forms_report r
+                left join actstream_action act on concat('Email sent: ''CRT - Constant Writer'' to ', r.contact_email, ' via govDelivery TMS') = act.description
+                WHERE act.verb = 'Contacted complainant:'
+                AND r.contact_email IN %s
+                order by act.timestamp desc
```
okkays, line 208: SQL readability nit: be consistent about upper vs lowercase keywords. I personally prefer uppercase.

```suggestion
                """  SELECT act.description AS description, r.contact_email AS email
                FROM cts_forms_report r
                LEFT JOIN actstream_action act ON CONCAT('Email sent: ''CRT - Constant Writer'' to ', r.contact_email, ' via govDelivery TMS') = act.description
                WHERE act.verb = 'Contacted complainant:'
                AND r.contact_email IN %s
                ORDER BY act.timestamp DESC
```

crt_portal/cts_forms/views.py
```diff
@@ -182,6 +183,38 @@ def _related_reports_count(report):
...
+            report.constant_writer = True
+
+
+def _was_sent_constant_writer_email(reports):
+    report_emails = tuple(map(_get_email, reports))
+    if len(report_emails) == 0:
+        return
+    with connection.cursor() as cursor:
+        try:
+            cursor.execute(
```
okkays, line 202: performance question: did you `EXPLAIN` this query to confirm it's leveraging your new index?

crt_portal/cts_forms/views.py
```diff
@@ -182,6 +183,38 @@ def _related_reports_count(report):
...
+
+
+def _set_constant_writer(contact_templates, report):
+    for contact_template in contact_templates:
+        if contact_template[1].upper() == report.contact_email.upper():
+            report.constant_writer = True
+
+
+def _was_sent_constant_writer_email(reports):
+    report_emails = tuple(map(_get_email, reports))
```
okkays, line 197:
```suggestion
    report_emails = tuple(_get_email(r) for r in reports))
```

style nit: using the more "pythonic" comprehensions (list, set, and dict) is generally preferred over map, filter, etc

though I really appreciate the functional approach!

https://stackoverflow.com/a/1247490/13231999

crt_portal/cts_forms/views.py
```diff
@@ -182,6 +183,38 @@ def _related_reports_count(report):
...
 
 
+def _get_email(report):
+    return report.contact_email
+
+
+def _set_constant_writer(contact_templates, report):
+    for contact_template in contact_templates:
+        if contact_template[1].upper() == report.contact_email.upper():
+            report.constant_writer = True
```
okkays, line 193: I'm not sure how I feel about setting this directly on the report, since the value isn't persistent and shouldn't be saved. Especially in this case, as this field has some history

What do you think about:
- Returning a `dict` of report_ids instead of setting values on the report itself
- Using the letter titles/etc (rather than a boolean)

crt_portal/cts_forms/views.py
```diff
@@ -182,6 +183,38 @@ def _related_reports_count(report):
         return 1
 
 
+def _get_email(report):
+    return report.contact_email
```
okkays, line 187: I'm not sure this function adds much for readability. Maybe just directly access contact_email, instead?

If this was for the `map` function, then the comprehension should help in this particular case. In other cases a lambda would also work:

e.g., if for some reason a comprehension wasn't available:

```py
map(lambda report: report.contact_email, reports)
```

crt_portal/cts_forms/views.py
```diff
@@ -271,6 +304,8 @@ def index_view(request):
         lambda text: Campaign.objects.filter(internal_name=text).first()
     )
 
+    _was_sent_constant_writer_email(requested_reports)
+
     final_data = {
         'form': Filters(request.GET),
         'profile_form': profile_form,
```
okkays, line 311: to further illustrate the comment I made above, the usage here would be:

```suggestion
    final_data = {
        'form': Filters(request.GET),
        'profile_form': profile_form,
        'sent_templates': _get_sent_templates_for_reports(requested_reports),
```

and then you'd render from that dict based on the report id in the html.


A second thought: This approach also makes the list page dependent on the constant writer icon query finishing. You might want to unblock that using some javascript after the page loads, [similar to this pattern](https://github.com/usdoj-crt/crt-portal/blob/develop/crt_portal/static/js/show_count.js)

crt_portal/static/sass/styles.scss
```diff
@@ -48,4 +48,4 @@
 @forward "custom/links";
 @forward "custom/modal";
 @forward "custom/print";
-@forward "custom/error";
+@forward "custom/error";
```
okkays, line 51: supernit: [configure your editor to add newlines at EOF](https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files)

okkays (changes requested):
Cool to see this working - extra exciting that you were able to do this efficiently without requiring a view!

crt_portal/cts_forms/views.py
ilodidoj: @kewitham @okkays we may need some place to save 'constant writer' for the future export and for analytics. We probably need to choose a place that can scale for other tags later on e.g. 'repeated email'

crt_portal/cts_forms/views.py
kewitham: Updated to use JS to run this query after page load

crt_portal/cts_forms/views.py
kewitham: I did and it wasn't using it but I think that's because I don't have enough data on local to make it worth it to use the index.

okkays: Here's what I'm seeing on local:

- 🏁 it appears to be using the description index
- 🐌 but it is taking a bit to scan the report emails for the `contact_email IN` filter

My suggestion: Add an additional index on email (which I imagine won't just be used here) seems to speed things up a bit:

```sql
CREATE INDEX IF NOT EXISTS cts_forms_report_contact_email_idx
    ON public.cts_forms_report USING btree
    (contact_email)
```

---

Proofs:

Without email index:

```sql
 Sort  (cost=1558.61..1558.85 rows=99 width=109)
   Sort Key: act."timestamp" DESC
   ->  Nested Loop  (cost=0.41..1555.32 rows=99 width=109)
         ->  Seq Scan on cts_forms_report r  (cost=0.00..1074.08 rows=16 width=22)
               Filter: ((contact_email)::text = ANY ('{caitlynboyd@example.com,loriglover@example.com,millersandra@example.com,ybanks@example.com,martinezrobert@example.com,ugarcia@example.com,stephanie86@example.com,jhawkins@example.com,williamsims@example.com,richardsonkeith@example.com,amycopeland@example.com,phogan@example.com,jenniferfox@example.com,pruittmelissa@example.com,richardcarroll@example.com}'::text[]))
         ->  Index Scan using actstream_action_description_idx on actstream_action act  (cost=0.41..30.02 rows=6 width=87)
               Index Cond: (description = concat('Email sent: ''CRT - Constant Writer'' to ', r.contact_email, ' via govDelivery TMS'))
               Filter: ((verb)::text = 'Contacted complainant:'::text)
```

image

With email index:

```sql
 Sort  (cost=545.03..545.28 rows=99 width=109)
   Sort Key: act."timestamp" DESC
   ->  Nested Loop  (cost=0.70..541.75 rows=99 width=109)
         ->  Index Only Scan using cts_forms_report_contact_email_idx on cts_forms_report r  (cost=0.28..60.51 rows=16 width=22)
               Index Cond: (contact_email = ANY ('{caitlynboyd@example.com,loriglover@example.com,millersandra@example.com,ybanks@example.com,martinezrobert@example.com,ugarcia@example.com,stephanie86@example.com,jhawkins@example.com,williamsims@example.com,richardsonkeith@example.com,amycopeland@example.com,phogan@example.com,jenniferfox@example.com,pruittmelissa@example.com,richardcarroll@example.com}'::text[]))
         ->  Index Scan using actstream_action_description_idx on actstream_action act  (cost=0.41..30.02 rows=6 width=87)
               Index Cond: (description = concat('Email sent: ''CRT - Constant Writer'' to ', r.contact_email, ' via govDelivery TMS'))
               Filter: ((verb)::text = 'Contacted complainant:'::text)
(8 rows)
```

image

okkays (changes requested):
Looking good, mostly style nits and one easy performance change.

okkays:
Oh, make sure to resolve conflicts

crt_portal/static/js/constant_writer.js
```diff
@@ -0,0 +1,33 @@
...
+      const target = Array.from(targets).find(target => target.getAttribute('data-id') === id);
+      target.classList.remove('hidden');
+    }
+  }
+
+  function makeQuery() {
+    const targets = dom.querySelectorAll('.show-cw');
+    const report_objects = {};
+    targets.forEach(target => {
+      report_objects[target.dataset.id] = target.getAttribute('data-email');
```
okkays, line 14: sometimes data-email is blank:

```
data-email="{{ datum.report.contact_email|default:'' }}">
```

does this code handle that properly? Would an `if (!getAttribute(data-email)) return)` make sense here?

okkays (approved):
One last comment - also, looks like you need to regen your migrations from that test failure

Pending that, LGTM!

PR #1322, opened 2023-02-07
Allow searching only for exact matches

okkays:
## What does this change?

Allow filtering personal description by exact match.

- 🌎 The personal description filter is actually a search. Typing "this is a test" into the box will also match "this is a fun test", "oh no this is a test", etc. That search doesn't have an "exact full match" functionality (the quote behavior doesn't behave that way, and doesn't provide for _full_ matches, anyway)
- ⛔ That's not going to work for the exact match count/link on the detail page. We only want to show _exact_ matches, there.
- ✅ This commit introduces a way to ask for "exact" matches, documents it, and uses it on the filter page.

Demo showing the issue and solve

image

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/filters.py
+++ b/crt_portal/cts_forms/filters.py
@@ -131,7 +131,11 @@ def report_filter(querydict):
                 kwargs[field] = querydict.getlist(field)
             elif field_options == 'violation_summary':
                 search_query = querydict.getlist(field)[0]
-                qs = qs.filter(violation_summary_search_vector=_make_search_query(search_query))
+                if search_query.startswith('^') and search_query.endswith('$'):
+                    # Allow for "exact match" using the common regex syntax.
+                    qs = qs.filter(violation_summary=search_query[1:-1])
+                else:
+                    qs = qs.filter(violation_summary_search_vector=_make_search_query(search_query))
             elif field_options == 'contact_phone':
                 # Removes all non digit characters, then breaks the number into blocks to search individually
                 # EG (123) 456-7890 will search to see if  "123" AND "456" AND "7890" are in the number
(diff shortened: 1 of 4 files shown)


crt_portal/cts_forms/filters.py
```diff
@@ -131,7 +131,11 @@ def report_filter(querydict):
                 kwargs[field] = querydict.getlist(field)
             elif field_options == 'violation_summary':
```
kewitham, line 132: It feels like this would be a good place for a switch statement - unrelated to this PR but maybe something we can refactor separately

kewitham (approved):
lgtm!

crt_portal/cts_forms/filters.py
okkays: There's some fun python history behind this - Python didn't introduce a switch until 3.10, and it's not really a switch - it's pattern matching that [can be very complex if you feel like giving your reviewer a headache](https://towardsdatascience.com/the-match-case-in-python-3-10-is-not-that-simple-f65b350bb025).

https://stackoverflow.com/a/11479840/13231999

Plus, [the official docs](https://docs.python.org/3/tutorial/controlflow.html) still recommend elif:

image

Even in this case, it's important to note that the `if..elif` is more flexible than `match` when you're not comparing the _same value_ to several constants - for example, there's a condition (`if date in field`) in this example that would throw things off:

image

okkays: all of that said, I agree, this could be refactored. Maybe into a helper function so we can return early and reduce nesting, e.g.:

```py
def build_filter(...):
  if field_options == 'etc':
    return qs.filter(...)
  if field_options == 'etc':
    return qs.filter(...)
  if field_options == 'etc':
    return qs.filter(...)  
  if field_options == 'etc':
    return qs.filter(...)


...


qs = build_filter(...)
```

kewitham: Interesting background. Makes sense - I wonder if we could just use `case create_date_start | create_date_end:` to solve that condition problem. I think it could be helpful for readability but that's just personal preference.

okkays:
Updated search helper text:

image

ilodidoj (approved):
LGTM!! @okkays ready to be merged in dev.

PR #1367, opened 2023-04-03
Fix bug with links containing underscores in response template admin panel

kewitham:
## What does this change?

Currently there is a bug in the admin panel where the code that converts html pasted response templates into markdown escapes underscores thereby breaking links with underscores. This PR adds logic to replace underscores in links before processing the html and then putting them back from the markdown conversion is done.

## Checklist:

Navigate to /admin and add a link with an underscore in it to the one of the response templates. Verify that the link is correctly converted to markdown and the underscore is not escaped.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/static/js/response_template_preview.js
+++ b/crt_portal/static/js/response_template_preview.js
@@ -31,10 +31,12 @@
     const turndown = new TurndownService();
     let pastedHtml = event.clipboardData.getData('text/html');
     // Turndown escapes underscores so maintain them in links by replacing them before processing
-    const aTags = pastedHtml.match(/(<[Aa]\s(.*)<\/[Aa]>)/g);
+    const parser = new DOMParser();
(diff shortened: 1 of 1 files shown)


okkays (changes requested):
[Using regex to parse HTML is scary](https://stackoverflow.com/a/1732454/13231999)

Here's an example of where this will behave strangely:

```html
<a href="_leave me">_leave me too_</a> _escape this!_ <a href="_leave me">leave _me_ three</a>
```
Produces:

image

But if done without the second link:

```html
<a href="_leave me">_leave me too_</a> _escape this!_
```

produces:

image

That is, your regex will capture and unescape anything between the first and last links in the document. You could use a non-greedy capture to try to get around this, but then you'll still be capturing the full anchor tag (and not just the "href")

okkays (approved):
👍 nice

PR #1383, opened 2023-04-11
Add Feature Flags to the application

okkays:
## What does this change?

- 🌎 It's currently not possible to turn on/off features without a code
  change.
- ⛔ This is dangerous, because we can't be sure how certain features will
  function against production data.
- ✅ This commit adds a "Feature Flags" module to the django app, which
  can be turned on and off from the admin panel
- 🔮 This currently turns on/off the entire feature for everybody. Future commits can turn features on/off for specific users.

Here's an example of adding/removing an element from an api response with the python functionality:

python

Here's a django template:

template

Here's an example of the available javascript functionality via the chrome console:

javascript

And here's a chrome inspector CSS example:

scss

### Author

### Reviewer

## Notes for reviewer:


crt_portal/features/templatetags/__init__.py
kewitham, line 1: is this intentional?

okkays: Yep. These files designate these folders as python modules

https://docs.python.org/3/tutorial/modules.html#packages

ilodidoj (approved):
@okkays merge this after our dev pairing this morning.

PR #1387, opened 2023-04-19
Add ICM Filter to the View All Table

okkays:
## What does this change?

- 🌎 Users can currently associate DJ numbers with reports.
- ⛔ They can't later find reports based on those numbers, though
- ✅ This commit adds a DJ number filter, which allows users to filter

Note from issue:

> should we add the same error validation for the sequence number field? (Error message if a non-numerical character is entered in this field)

Rather than add validation, we'll just prevent them from entering it :D

Filter look and feel (plus screenreader):

filter

### Author

### Reviewer

## Notes for reviewer:


crt_portal/static/js/complaint_view_filters.js
```diff
@@ -292,6 +303,26 @@
     }
   };
 
+  /** Produces a new multi-field value given one of its constituent parts. */
+  function buildMultiValue(target, newValue) {
+    const fieldName = target
```
kewitham, line 308: This breaks with per page and grouping because those fields don't have ids

kewitham (changes requested):
just one comment but icm filter functionality looks great!

crt_portal/static/js/complaint_view_filters.js
okkays: I think I've addressed this; can you take another look? (If you're still seeing the issue, can give you me steps to reproduce?)

ilodidoj (approved):
@okkays passed local docker test.  If @kewitham comments, concerns resolved, this can be merged to dev.

kewitham (approved):

PR #1392, opened 2023-04-24
Fix Long URLs when grouping by long descriptions

okkays:
Basically, replaces "violation_summary=report.description" with "violation_summary=report.id" for grouping mode.

This adds a lookup on the report id when filtering, but drastically reduces the length of the URL.

## What does this change?

- 🌎 We currently pass the full description as a filter when grouping.
- ⛔ This shows up in the URL, which causes URLs to be too long sometimes (greater than 8000 characters).
- ✅ This commit uses an arbitrary report id as the filter criteria, instead, meaning that the URL will stay small at the cost of one report lookup.

GIF demonstrating that the URL is smaller now

### Author

### Reviewer

## Notes for reviewer:


ilodidoj (commented):
@okkays please check the conflict on this branch.  I can review again tomorrow.

kewitham:
The violation summary in the url is used on the individual report pages to reconstruct the query and allow users to page through all reports in the group. That is usually found here but now it's missing:
image

kewitham:
It looks like this value isn't being pulled correctly - should say 17 in the case I was testing for the second button - 
image

kewitham (changes requested):
Unfortunately because of the way urls are used to rebuild queries this has some unexpected side effects.

okkays:
@kewitham I'm not able to reproduce those missing reports - what's the description you're trying to match against, and how many records have it?

Here's the feature working as expected:

descriptions-url

bulk actions

kewitham (approved):
lgtm!

PR #1396, opened 2023-05-01
Ignore browser cache when static files have changed

okkays:
This is a Draft PR as part of the spike to explore how we might handle force-reloading of static files. We'll discuss this approach and modify/merge as appropriate.

## What does this change?

- 🌎 Users aren't refreshing the page, and so are getting cached static files.
- ⛔ This is bad, because our python often depends on our javascript or CSS having the latest versions. When they are out of sync, unexpected behavior can happen.
- ✅ This commit extends the static tag to use the current directory's modified time as a way to indicate when the application was last deployed. This means that the browser will request new versions of files when the application is deployed or restaged (whenever we've stopped just serving the "same" code).

I've also added this new static tag to the builtin template context in settings.py, so we don't need to `{% load static %}` in every file anymore.

Here's the network tab showing the values loading without 

image

### Author

### Reviewer

## Notes for reviewer:

To bump the timestamp when testing locally, running `touch .` from the repository root will cause the package-lock's updated time to be changed, simulating the content being re-cloned.

### DIFF
--- a/crt_portal/api/views.py
+++ b/crt_portal/api/views.py
@@ -133,7 +133,6 @@ class ResponseTemplatePreviewBase:
 
     def _add_css(self, body):
         return body + mark_safe(
-            '{% load static %}'
             '<link rel="stylesheet" href="{% static "css/compiled/template-preview.css" %}">'
         )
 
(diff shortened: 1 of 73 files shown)


kewitham (approved):
lgtm!

PR #1404, opened 2023-05-10
Kk/standalone docker

okkays:
## What does this change?

Explicitly, this should **not** affect the current `docker compose up/run/down` usage. That should function as it previously did.

The goal here is to be able to run other code (to review other branches, etc) in a sterile way without affecting the development docker.

- 🌎 Our docker-compose depends on local code, volumes, and ports being available. This is good for development mode, because it means we can sync our local changes with a container.
- ⛔ It's bad for running tests or other processes locally, though, because we can't modify the code until the tests are finished running.
  which means that we can't
- ✅ This commit separates the concerns, with a "docker-compose" file to contain any common attributes of containers, and override files for tests and local development. Local development is the default override file, so `docker compose up`, etc, should continue to function as normal.
- 🔮 Future commits will make use of this standalone docker-compose for running tests and other ci/cd concerns locally, and could be used to run and switch between multiple versions of the codebase at once.

Here's an example of spinning up four different simultaneous docker instances on different ports:

four-opt

Here's how multiple container groups look in docker desktop:

image

### Author

### Reviewer

## Notes for reviewer:


ilodidoj (approved):
@okkays validated locally.

PR #1447, opened 2023-06-27
Add new api route for referral flow

kewitham:
[Issue 1577](https://github.com/usdoj-crt/crt-portal-management/issues/1577)

## What does this change?
This PR adds an API route for the new referral flow so users can send, print, or copy response templates without refreshing the page.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/api/views.py
+++ b/crt_portal/api/views.py
@@ -305,12 +305,7 @@ class ReferralResponse(APIView):
         action = request.data['action']
         preview = {'preview': {}}
         if not action:
-            return JsonResponse(
-                {
-                    'response': 'No action provided',
-                    **preview,
-                  }, status=400
-            )
+            return JsonResponse({'response': 'No action provided', **preview}, status=400)
         if action == 'send':
             try:
                 email_response = crt_send_mail(report, template)
--- a/crt_portal/api/views.py
+++ b/crt_portal/api/views.py
@@ -2,13 +2,13 @@ from api.filters import form_letters_filter, reports_accessed_filter, autorespon
 from django.utils.html import mark_safe
 from api.serializers import ReportSerializer, ResponseTemplateSerializer, RelatedReportSerializer
 from cts_forms.filters import report_filter
-from cts_forms.mail import CustomHTMLExtension
+from cts_forms.mail import CustomHTMLExtension, crt_send_mail
 from cts_forms.models import Report, ResponseTemplate
 from cts_forms.views import mark_report_as_viewed, mark_reports_as_viewed
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.http import HttpResponse
-from django.shortcuts import render
+from django.shortcuts import render, get_object_or_404
 from django.template import Context, Template
 from rest_framework import generics
 from rest_framework import permissions
@@ -39,7 +39,8 @@ def api_root(request, format=None):
         'report-count': reverse('api:report-count', request=request, format=format),
         'related-reports': reverse('api:related-reports', request=request, format=format),
         'form-letters': reverse('api:form-letters', request=request, format=format),
-        'report-cws': reverse('api:report-cws', request=request, format=format)
+        'report-cws': reverse('api:report-cws', request=request, format=format),
+        'referral-response': reverse('api:referral-response', request=request, format=format)
     })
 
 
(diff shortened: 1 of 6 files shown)


crt_portal/api/views.py
```diff
@@ -284,3 +285,32 @@ def get(self, request):
...
+        preview = {}
+        if action == 'send':
+            try:
+                email_response = crt_send_mail(report, template)
+                if email_response:
+                    response = f'Sent referral email template #{template.id} to report #{report.id}'
+                else:
+                    response = f'Referral email template #{template.id} failed to send to report #{report.id}'
+            except Exception as e:
+                response = f'Referral email template #{template.id} failed to send to report #{report.id}: {e}'
```
okkays, line 315: this try-except is a bit broad (nothing is ever going to go wrong with the `if email_response else` portion unless there's a syntax error in the code, in which case you _don't_ want to catch the exception.

These kinds of nested blocks can be simplified a bit if you're able to return early. For example, maybe instead:

```py
if action == 'send':
  response = {'preview': preview}
  try:
    email_response = crt_send_mail(report, template)
  except Exception as e:
    return Response({'email_response': f'Referral email template #{template.id} failed to send to report #{report.id}: {e}', **response})
  if not email_response:
    return Response({'email_response': f'Referral email template #{template.id} failed to send to report #{report.id}', **response})
  return Response({'email_response': f'Sent referral email template #{template.id} to report #{report.id}', **response})
```

You could further reduce nesting (especially if 'copy' and 'print' turn out to be complicated) by making a helper function for the `if action == 'send'` portion

okkays (changes requested):

crt_portal/api/views.py
```diff
@@ -284,3 +285,32 @@ def get(self, request):
...
+        template = get_object_or_404(ResponseTemplate, pk=template_id)
+        action = request.data['action']
+        preview = {}
+        if action == 'send':
+            try:
+                email_response = crt_send_mail(report, template)
+                if email_response:
+                    response = f'Sent referral email template #{template.id} to report #{report.id}'
+                else:
+                    response = f'Referral email template #{template.id} failed to send to report #{report.id}'
```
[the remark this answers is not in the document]
ilodidoj: @kewitham @okkays we do get reasons for failure some times, it's the error field under TMS Emails > Error message field. But I think this happen later once we pass the msg to TMS client.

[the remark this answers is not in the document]
kewitham: crt_send_mail function calls the [send_mail function](https://docs.djangoproject.com/en/4.2/topics/email/) which only returns 0 or 1 depending on how many messages are successful so I'm not sure how we would include the additional information.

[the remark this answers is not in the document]
okkays: Sounds good. Then I'd just return a `502` if something goes wrong (indicating the mail service failed to send)

crt_portal/api/views.py
```diff
@@ -284,3 +285,36 @@ def get(self, request):
...
+
+    def post(self, request):
+        report_id = request.data['report_id']
+        report = get_object_or_404(Report, pk=report_id)
+        template_id = request.data['template_id']
+        template = get_object_or_404(ResponseTemplate, pk=template_id)
+        action = request.data['action']
+        preview = {'preview': {}}
+        if not action:
+            return Response({'response': '500: No action provided', **preview})
```
okkays, line 308: I don't think this actually returns a 500

also - 500 should be in the event of a unexpected server error (our code broke) this should probably be 400

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors

https://docs.djangoproject.com/en/4.2/ref/request-response/#jsonresponse-objects

```suggestion
            return JsonResponse(
              {
                'message': 'No action provided',
                **preview,
              }, status=400
              )
```

okkays (changes requested):

crt_portal/api/views.py
```diff
@@ -284,3 +285,41 @@ def get(self, request):
...
+              {
+                'message': 'No action provided',
+                **preview,
+              }, status=400
+              )
+        if action == 'send':
+            try:
+                email_response = crt_send_mail(report, template)
+            except Exception as e:
+                return Response({'response': f'502: Referral email template #{template.id} failed to send to report #{report.id}: {e}', **preview})
```
okkays, line 318: Sorry, here as well

crt_portal/api/views.py
```diff
@@ -284,3 +285,41 @@ def get(self, request):
...
+            if not email_response:
+                return Response({'response': f'502: Referral email template #{template.id} failed to send to report #{report.id}', **preview})
```
okkays, line 320: - And here

okkays (changes requested):
sorry for the extra back and forth, lgtm after this

okkays (approved):

okkays:
(Approved pending test updates to match the new statuses!)

ilodidoj:
@kewitham @okkays Probably need to fix the unit test issue before merging, CI/CD failure needs to be resolved.

PR #1453, opened 2023-06-30
Report detail testing

kewitham:
[Issue 1583](https://github.com/usdoj-crt/crt-portal-management/issues/1583)

## What does this change?

This PR fixes a bug with the back to all button where clicking it from a report detail page always takes users back to the dashboard/activity/ page. It also updates the logic for the report detail pagination to account for the different query type when navigating from the activity dashboard.

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/tests/integration-authed/report_detail.py
```diff
@@ -0,0 +1,22 @@
...
+
+    page.goto("/form/view")
+
+    with page.expect_navigation():
+        page.evaluate("document.querySelector('.td-link').click()")
+
+    assert page.is_visible("#contact-info")
+
+    with page.expect_navigation():
+        page.evaluate("document.getElementById('backtoall').click()")
```
okkays, line 20: I've found that in most cases, you can use the `click()` method on locator directly.

Also, to avoid assigning ids to things, you can use locator().filter() - for example:

```
    page.locator('.crt-landing--example_option').filter(has_text="Victim of a hate crime").click()

    assert page.locator('p').filter(has_text="Attacks, threats of violence, or destruction of property at place of worship").is_visible()
    assert page.locator('a').filter(has_text="Get help for hate crimes").is_visible()
```

crt_portal/cts_forms/tests/integration-authed/report_detail.py
```diff
@@ -0,0 +1,22 @@
...
+
+    login_as_superuser(page)
+
+    page.goto("/form/view")
+
+    with page.expect_navigation():
+        page.evaluate("document.querySelector('.td-link').click()")
+
+    assert page.is_visible("#contact-info")
+
```
okkays, line 18: This asserts that the correct kind of page has loaded, but I think this might still pass without your bug fix

It'd be cool to see this test:
1. Take note of what and how many reports are displayed on the list page.
2. Assert that the clicked report loads the correct ID
3. Assert that the next / prev is showing "1 of correct number"
4. Click the next arrow a few (for example, 3) times
5. Click the back arrow a few (for example, 2) times
6. Assert that the pagination is correct (for example, "2 of correct number"
7. Click back to all
8. Assert that we've gone back to the correct place.

crt_portal/cts_forms/views.py
```diff
@@ -155,6 +153,33 @@ def setup_filter_parameters(report, querydict):
...
 
+def get_requested_ids(return_url_args):
+    return_url_querydict = QueryDict(return_url_args[1:])
+    activity = return_url_querydict.get('activity', None)
+    if activity:
+        requested_query = reconstruct_activity_query(return_url_args)
+        return list(map(int, requested_query.values_list('target_object_id', flat=True)))
+    else:
+        requested_query = reconstruct_query(return_url_args)
+        return list(requested_query.values_list('id', flat=True))
```
okkays, line 164:
```suggestion
    requested_query = reconstruct_query(return_url_args)
    return list(requested_query.values_list('id', flat=True))
```

supernit: unecessary else after if + return; since this is your base case (standard report list), imo it shouldn't be indented

crt_portal/cts_forms/views.py
```diff
@@ -155,6 +153,33 @@ def setup_filter_parameters(report, querydict):
...
+    will return all records.
+    """
+    querydict = QueryDict(next_qp)
+
+    _, selected_actions = dashboard_filter(querydict)
+    sort_expr, _ = activity_sort(querydict.getlist('sort'))
+    if selected_actions != []:
+        selected_actions = selected_actions.order_by(*sort_expr)
+
+    return selected_actions
```
okkays, line 180: If I understand what's going on here, it looks like you're handling an edge case where selected actions _is_ empty?

If so, this approach might be more extensible in the case that you have more conditions / want to do more than just order later on:

```suggestion
    if not selected_actions:
        return selected_actions

    return selected_actions.order_by(*sort_expr)
```

okkays (changes requested):

crt_portal/cts_forms/tests/integration-authed/report_detail.py
kewitham: I did most of this but the test was timing out when I added all of it. I think I can use pytest-timeout to set a longer timeout for the test but I wasn't sure if that was a good idea.

okkays (changes requested):
CI/CD is failing due to a whitespace issue.

If you don't already have this enabled, it's really useful: https://stackoverflow.com/a/30884298/13231999

.circleci/config.yml
```diff
@@ -134,7 +138,8 @@ jobs:
           environment:
             PUPPETEER_EXECUTABLE_PATH: /usr/bin/google-chrome
           command: pipenv run pytest crt_portal/cts_forms/tests/integration-authed/*.py --base-url=http://localhost:8000
-
+      - store_artifacts:
+          path: ./e2e-screenshots
```
okkays, line 142: yessssssssssss

okkays (approved):
very nice! 👏

PR #1458, opened 2023-07-06
Back to all and GA fix

kewitham:
## What does this change?

This PR fixes the back to all button so it directs user to the appropriate page from the individual report page rather than always sending them to the activity dashboard page. An e2e test for the report detail page will be added in another PR.
It also fixes a js error for the GA script.

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/features/templatetags/feature_script.py
+++ b/crt_portal/features/templatetags/feature_script.py
@@ -10,7 +10,7 @@ register = template.Library()
 
 
 @register.simple_tag
-def feature_script():
+def feature_script(csp_nonce):
     all_features = Feature.objects.all()
 
     feature_json = json.dumps({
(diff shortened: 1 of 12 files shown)


crt_portal/features/templatetags/feature_script.py
```diff
@@ -23,7 +23,7 @@ def feature_script():
     ], separators=(',', ':'))
 
     script = textwrap.dedent(f"""
-        <script>
+        <script nonce="{{ request.csp_nonce }}">
```
okkays, line 26: just noting after talking that this needs to be fixed, so I can be notified when you re-request review 😅

okkays (changes requested):

okkays (approved):

crt_portal/features/templatetags/feature_script.py
kewitham: updated! it looks like you don't actually see the csp_nonce locally though (I checked some existing instances) so you just see <script nonce>

PR #1466, opened 2023-07-11
Fix TMS admin template.

okkays:
## What does this change?

- 🌎 The TMS admin view shows admin details
- ⛔ Right now it's not, though, because it's using the wrong template.
- ✅ This commit fixes the template
- 🔮 Future commits will provide any additional fixes that are needed once we see how this behaves in prod.

image

### Author

### Reviewer

## Notes for reviewer:


kewitham (changes requested):
Do we still need to use email.html in the api views.py and mail.py files?

okkays:
> Do we still need to use email.html in the api views.py and mail.py files?

Yes - that's a different `email.html`. Previously we had both:

```
crt_portal/cts_forms/templates/email.html
crt_portal/templates/email.html
```

Django could only ever find one of them, which caused the bug that this PR fixes.

ilodidoj:
> > Do we still need to use email.html in the api views.py and mail.py files?
> 
> Yes - that's a different `email.html`. Previously we had both:
> 
> ```
> crt_portal/cts_forms/templates/email.html
> crt_portal/templates/email.html
> ```
> 
> Django could only ever find one of them, which caused the bug that this PR fixes.

@okkays @kewitham If we currently using HTML version for auto reply, then, please check if any impact with this changes and update both auto reply and manual HTML reply accordingly.

okkays:
@ilodidoj @kewitham these changes only affect the TMS admin page template.

They don't affect user-facing emails.

The two templates had the same name which was confusing to both us and Django - but they take different context variables and are used in separate views. This renames the TMS admin template, and leaves the user email template alone.

PR #1499, opened 2023-08-11
Referrals: Enable the Print button from the modal

okkays:
[Issue link](https://github.com/usdoj-crt/crt-portal-management/issues/1590)

## What does this change?

- 🌎 The referral modal lets users send and print referrals.
- ⛔ Right now the print button does nothing!
- ✅ This commit generates printed documents from the server and triggers a browser download on the client.

referral4

### Author

### Reviewer

## Notes for reviewer:


kewitham (changes requested):
Not a huge deal but is there any way to specify page breaks so there's not awkward breaks like this?
image

okkays:
> Not a huge deal but is there any way to specify page breaks so there's not awkward breaks like this?

Good point - we should actually be able to specify header and footer content for the PDF creator. It's kind of tricky, though (we'd need to break apart the HTML template into header / content / footer), so I might add it as a followup ticket.

kewitham (approved):

PR #1583, opened 2023-11-02
What's Happening Where Enhancements

kewitham:
[What's Happening Where/Incident Location enhancements#1654](https://github.com/usdoj-crt/crt-portal-management/issues/1654)

## What does this change?

This updates the What's Happening Where and Incident Location data to be filterable by city and district respectively.

image

image

### Author

### Reviewer

## Notes for reviewer:


okkays (changes requested):
This is so cool! I can't wait to see this against production.

Can you run `dump_dashboards_from_db` to update dashboards.json? (This will make your additions persistent)

```docker-compose run web python /code/crt_portal/manage.py dump_dashboards_from_db```

File: crt_portal/static/js/data-filters.js [hunk 1]
```diff
@@ -88,8 +88,37 @@
     document.querySelector('.data-content').hidden = false;
   }
 
+  function initCityFilter(filter) {
+    filter?.addEventListener('change', () => {
+      filterByCity(filter.value);
```
okkays, line 93: WDYT (might be a good candidate for a stacked or followup PR even) about making these shareable via setParam() / getParam()

jupyterhub/assignments/intake-dashboard/report_districts.ipynb
```diff
@@ -0,0 +1,139 @@
...
+    "    'commercial_or_public': 'Discriminated against in a commercial location or public place',\n",
+    "    'something_else': 'Something else happened',\n",
+    "}\n",
+    "readable_start = start.strftime(\"%B %d, %Y\")\n",
+    "readable_end = end.strftime(\"%B %d, %Y\")\n",
+    "report_district_data = report_districts.DataFrame()\n",
+    "districts = report_district_data['district'].values\n",
+    "data_divs = []\n",
+    "district_dropdown = []\n",
+    "for i, district in enumerate(districts):\n",
```
okkays, line 71: A couple of (optional) tips in the vein of functional, immutable code:
- Refactor the count controlled loop using `zip`
- Use a helper function to build the data_divs list with a comprehension instead of `append`

you might be able to use `zip` instead of `enumerate` to avoid the need for a counter here.

```py
complaints = [
  primary_complaint_choices.get(complaint_choice)
  for complaint_choice
  in report_district_data['complaint'].values
]
classifications = report_district_data['classification'].values
states = report_district_data['state'].values
counts = report_district_data['count'].values
records = zip(districts, complaints, classifications, states, counts)
data_divs = [
  make_data_div(district, complaint, classification, state, count)
  for district, complaint, classification, state, count in records
]

def make_data_div(district, complaint, classification, state, count):
  return f"""that multiline string"""


```

okkays: (this pattern can also apply to the other file)

jupyterhub/assignments/intake-dashboard/report_districts.ipynb
```diff
@@ -0,0 +1,139 @@
...
+    "    r.district as district,\n",
+    "    r.primary_statute as classification,\n",
+    "    r.primary_complaint as complaint,\n",
+    "    r.location_state as state,\n",
+    "    COUNT(*) count\n",
+    "FROM cts_forms_report r\n",
+    "WHERE district notnull AND primary_complaint notnull\n",
+    "GROUP BY complaint, district, classification, state\n",
+    "HAVING COUNT(*) > 1\n",
+    "ORDER BY count desc"
```
okkays, line 45: this (and your other query) should be very efficient, nice work!

Only suggestion - do you want to filter by the `start` and `end` date filters from above?

okkays (changes requested):

okkays:
one more comment around accessibility, sorry

it might be more accessible to render this as a table, so that screen readers interact with it appropriately (for example, by indicating what column the reader is on as they peruse records further down the list)

this would also allow you to hide the final row divider using css rather than `hr`

okkays (approved):
LGTM! :D

randyabramson:
@kewitham - just thinking about these reports again....what is the time range for these reports?

Are they starting with Aug. 9 and will just add daily totals each coming day?

And there's no way to change the date range in this MVP, right? cc @okkays

kewitham:
@randyabramson it's set to take the current time and go 90 days back from that. We could change the logic to grab data from a different range but users wouldn't be able to make changes in the UI to change the range dynamically.

randyabramson:
@kewitham  - thanks...I do see that date range in the UI.  Sorry I missed that!

PR #1594, opened 2023-11-20
Run less queries when exporting data

okkays:
https://github.com/usdoj-crt/crt-portal-management/issues/1659

## What does this change?

- 🌎 When we export data, we stream and export it in small chunks to avoid out-of-memory errors.
- ⛔ Right now, the chunks are too small. We're paginating at 2,000 records at a time. This means we'll never have RAM issues, but can mean a lot of SQL queries need to run to export relatively small sets. Queries have overhead, which makes exporting slow or even time out.
- ✅ This commit moves the balance towards more RAM / less time.
- 🔮 Future commits will ensure we're selecting only the values we need to, as well

N/A

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/crt_portal/settings.py
+++ b/crt_portal/crt_portal/settings.py
@@ -178,6 +178,11 @@ AUTH_PASSWORD_VALIDATORS = [
     },
 ]
 
+# Controls admin export page size.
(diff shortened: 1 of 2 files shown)


ilodidoj (approved):
LGTM

PR #1600, opened 2023-11-22
Add a simple model for Tags

okkays:
https://github.com/usdoj-crt/crt-portal-management/issues/1656

## What does this change?

- 🌎 We're adding tags to reports
- ⛔ We have no database backing for those, yet
- ✅ This commit adds a simple DB object and association on reports.
- 🔮 This is part of a larger tags MVP, future commits will use these models in the UI.

Admin panel fields

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/models.py
```diff
@@ -255,6 +255,23 @@ def __str__(self):
...
 
+class Tag(models.Model):
+
+    class Meta:
+        permissions = (
+            ("assign_tag", "Can assign tags to reports"),
+        )
+
+    name = models.CharField(max_length=255, null=False, blank=False, help_text="The name of the tag that will be shown throughout the app.")
+    section = models.TextField(choices=SECTION_CHOICES, null=True, blank=True, default=None, help_text="The section to which this tag applies. If set, this tag will only be available to reports in this section.")
```
kewitham, line 266: Do we want to allow tags to be used by multiple sections?

okkays: I think so, yeah. But in that case, I anticipated making the tag available to all sections (by setting section to `null`)

If we want to allow a tag to a few sections (but not all), it'd mean that we need a many-to-many relationship, which would mean we'd be executing a join every time we queried for tags based on section.

Probably not the worst, but it's additional complexity we can avoid and I'm not sure the business case of sharing tags between two sections is worth it.

okkays: (@blaughman curious to hear your take)

blaughman: @okkays, Amanda, @ilodidoj  and I discussed this during today's standup.  I think having the section acronym associated with the label would be helpful. As we saw with secondary review, universal labels may have different meanings with different sections. I don't want to recreate that same experience. I'm attaching a file SPL shared with me detailing their acronyms for labeling reports and the associated definitions.  Hope this helps and I hope all is going well for you. 
[SPL ISSUES - UPDATED.pdf](https://github.com/usdoj-crt/crt-portal/files/13504289/SPL.ISSUES.-.UPDATED.pdf)

kewitham (approved):

PR #1640, opened 2023-12-14
Add warning when complaint status is updated without saving

kewitham:
## What does this change?
This PR adds an alert when a user makes a change to the complaint actions card and doesn't save before trying to leave the page. Note that we cannot customize the alert because it's default to the browser.
Image
### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/static/js/save_warning.js
+++ b/crt_portal/static/js/save_warning.js
@@ -1,0 +1,10 @@
+(function() {
+    let formChanged = false;
+    const statusForm = document.querySelector("#complaint-view-actions");
+    statusForm.addEventListener('change', () => formChanged = true);
+    window.addEventListener('beforeunload', (e) => {
+      if (formChanged) {
+        e.returnValue = 'Changes you made may not be saved.';
+      }
+    });
+  })();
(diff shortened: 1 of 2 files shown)


okkays (changes requested):
- The popup appears when you click "Apply changes"; since they're saving, we want to let them navigate away
- This only applies to status; would this be useful for summary / other fields, as well?

crt_portal/static/js/save_warning.js
```diff
@@ -0,0 +1,16 @@
+(function() {
+  let showWarning = false;
+  const forms = document.querySelectorAll('.usa-form');
+  forms.forEach(form => {
+    form.addEventListener('change', () => (showWarning = true));
+  });
+  const btns = document.querySelectorAll('.usa-button');
```
okkays, line 7: usa-button might be overselecting, here.

uhoh

maybe something like "type=submit" would be a better selection?

okkays (changes requested):

okkays (approved):
lgtm! thanks for dealing with the back and forth!

PR #1682, opened 2024-01-19
Add sliders to control organization name search

okkays:
https://github.com/usdoj-crt/crt-portal-management/issues/1705

## What does this change?

- 🌎 We can fuzzy search with ~name right now
- ⛔ The search isn't configurable, which means three letter search terms get the same fuzziness as long ones
- ✅ This commit adds sliders to control the "percentag" of fuzziness applied to the search

Image

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/filters.py
```diff
@@ -310,28 +336,36 @@ class OctetLength(Func):
     output_field = IntegerField()
 
 
-def filter_by_similar(queryset, column, search_text):
+def filter_by_similar(queryset, column, *, query, soundslike, lookslike):
```
kewitham, line 339: I'm a little confused about how the "sounds like" and "looks like" variables impact each other. Like if I set the looks like filter to be too strict will I exclude some of the sounds like filtering? We should probably explain that a bit for the user on the front end too.

crt_portal/cts_forms/templates/forms/complaint_view/index/filters/incident_location_filter.html
```diff
@@ -1,18 +1,11 @@
...
-{% endif %}
-  <div id="incident_location-filter">
-    <label for="id_{{form.fields.location_name.widget.attrs.name}}">
-      {{ form.location_name }}
-    </label>
-  </div>
-{% if ENABLED_FEATURES.fuzzy_location_name %}
-</span>
-{% endif %}
+<div id="incident_location-filter">
```
kewitham, line 8: should we have some indication in the filter pills above the table that fuzziness is being used?

kewitham (commented):
a couple questions but this is really cool!

ilodidoj:
@okkays I have not had a chance to test this PR. Is the slider keyboard accessible? All input must be key board accessible for 508 and accessibility compliance.

okkays:
@ilodidoj They are! I'd tested that myself, but they're also uswds components so we can trust they've done their due diligence there, too

ilodidoj:
> @ilodidoj They are! I'd tested that myself, but they're also uswds components so we can trust they've done their due diligence there, too

Thanks for looking into it.  Trust but verify 😄

crt_portal/cts_forms/templates/forms/complaint_view/index/filters/incident_location_filter.html
okkays: hmm good point. To keep it simple, I might just add a suffix like "Organization name (fuzzy)" or "Organization name (exact)"

crt_portal/cts_forms/filters.py
okkays: they OR together, so they don't actually impact each other - if you add more sounds like, you get more sounds like. If you then adds more looks like, you get more looks like in the mix, too.

crt_portal/cts_forms/templates/forms/complaint_view/index/filters/incident_location_filter.html
okkays: decided not to keep it simple - here's what I've come up with:

Image
Image
Image

crt_portal/cts_forms/filters.py
okkays: I've updated the tooltips, plus the chips should help express this now, too

kewitham (approved):
lgtm!

PR #1690, opened 2024-01-24
Retention field bug fix

kewitham:
## What does this change?

This PR fixes a bug where the retention_schedule field registered a change when any change was made after it was set on the individual report page.

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/forms.py
```diff
@@ -1818,11 +1818,17 @@ def clean_dj_number(self):
         return dj_number
 
     def clean_retention_schedule(self):
-        if not self.field_changed('retention_schedule'):
-            return self.cleaned_data.get('retention_schedule')
+        old = self.initial.get('retention_schedule', None)
+        new = self.data.get('retention_schedule', None)
```
okkays, line 1822: from our conversation - I think using cleaned_data for `new` will give you a reference to the RetentionSchedule without having to fetch it yourself

okkays (changes requested):

crt_portal/cts_forms/forms.py
kewitham: Looked into this more and what I remembered was the cleaned_data value was actually the same as the initial value as if it didn't change. I think this is because the field is set to disabled when the user does not have permission to change it so [any changes to it are ignored](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#:~:text=The%20disabled%20boolean%20argument%2C%20when,from%20the%20form's%20initial%20data.). This would mean we don't need the clean_retention_schedule function and we shouldn't expect errors in the test.

okkays: nice - this is definitely why I put that check in there, too:

> Even if a user tampers with the field’s value submitted to the server, it will be ignored in favor of the value from the form’s initial data.

LGTM then!

okkays (approved):

PR #1731, opened 2024-03-07
Add pagination to disposition bulk actions

kewitham:
[Implement Disposition Batching / Approval: Intake Staff](https://github.com/usdoj-crt/crt-portal-management/issues/1696)

## What does this change?
This PR adds pagination to the reports on the disposition bulk actions page
Image
### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/cts_forms/views.py
+++ b/crt_portal/cts_forms/views.py
@@ -980,6 +980,12 @@ class DispositionActionsView(LoginRequiredMixin, FormView):
             })
         return data
 
+    def reconstruct_id_args(self, ids):
+        id_args = ''
+        for id in ids:
+            id_args += f'&id={id}'
+        return id_args
+
     def get(self, request, id=None):
         return_url_args = request.GET.get('next', '')
         return_url_args = urllib.parse.unquote(return_url_args)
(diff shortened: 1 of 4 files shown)


crt_portal/cts_forms/views.py
```diff
@@ -980,6 +980,12 @@ def get_report_data(self, requested_reports):
             })
         return data
 
+    def reconstruct_id_args(self, ids):
+        id_args = ''
+        for id in ids:
+            id_args += f'&id={id}'
+        return id_args
```
okkays, line 987: A _fun_ ctional way to do this:

```suggestion
        return ''.join([f'&id={id}' for id in ids])
```

crt_portal/cts_forms/views.py
```diff
@@ -988,41 +994,58 @@ def get(self, request, id=None):
...
         if selected_all:
             requested_query = reconstruct_query(query_string)
+            selected_report_args = 'all=all'
         else:
             requested_query = Report.objects.filter(pk__in=ids)
+            selected_report_args = self.reconstruct_id_args(ids)
 
-        data = self.get_report_data(requested_query)
+        if requested_query.count() > 500:
+            raise PermissionDenied
```
okkays, line 1003: is PermissionDenied right for this? maybe this is a BadRequest?

okkays (approved):
Really nice sized PR. Super understandable, thanks for keeping it small!

crt_portal/cts_forms/views.py
kewitham: I used PermissionDenied because that's what we use for bulk actions currently when the query count is too high. Looking into it it does seem like BadRequest is more appropriate. Wdyt about changing the exception for bulk actions as well?

okkays: I think that makes sense. I don't see anywhere in the frontend bulk actions that PermissionDenied is referenced (403, etc)

crt_portal/cts_forms/views.py
```diff
@@ -1035,8 +1055,14 @@ def post(self, request, id=None):
 
         if confirm_all:
             requested_query = reconstruct_query(query_string)
+            selected_report_args = 'all=all'
         else:
             requested_query = Report.objects.filter(pk__in=ids)
+            selected_report_args = self.reconstruct_id_args(ids)
+
+        if requested_query.count() > 500:
+            raise PermissionDenied
```
okkays, line 1064: should this be changed, too?

okkays (approved):

PR #1780, opened 2024-04-15
Adds a route for building a PDF form

okkays:
https://github.com/usdoj-crt/crt-portal-management/issues/1801

## What does this change?

This makes no user-facing changes, yet.

- 🌎 The report form is maintained as a static PDF.
- ⛔ This is hard to maintain and is bound to get out of date versus the web form
- ✅ This commit adds a route that generates a (unstyled, basic) printable form. This generation works with things like Django's translation functionality.
- 🔮 Future commits will:
    - style this form, including allowing a generated large-print version
    - add translations where they're missing
    - replace the existing form link with this

See http://localhost:8000/report/printable or http://localhost:8000/report/printable?lang=es

Example of a translated form. Note that some blocks still need translated content, because the web form doesn't have them:

- [file_a_complaint.pdf](https://github.com/usdoj-crt/crt-portal/files/14982173/file_a_complaint.pdf)
- [presentar_una_querella.pdf](https://github.com/usdoj-crt/crt-portal/files/14982174/presentar_una_querella.pdf)

### Author

### Reviewer

## Notes for reviewer:


kewitham (approved):
lgtm!

kewitham:
One nit - when I remove the lang parameter I would expect it to default back to English but it stays on the previous language (Spanish in this case).

okkays:
> when I remove the lang parameter I would expect it to default back to English but it stays on the previous language (Spanish in this case).

It remembers the language you were viewing with - this is the same mechanism as the language selector, which remembers what language you were viewing with (it's not specific to the PDF rendering)

image

PR #1915, opened 2024-09-13
Add search and filtering to resources table

kewitham:
[Resource Table Search Implementation #1057](https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/1057)

## What does this change?
This PR adds an api path to facilitate dynamic searching and filtering on the resources table.

Image

### Author

### Reviewer

## Notes for reviewer:


crt_portal/static/js/resources.js
```diff
@@ -91,21 +118,154 @@
...
+      setTimeout(() => {
+        makeQuery();
+      }, 1500);
     };
-  });
+
+    const tagInput = document.querySelector('#id_tags-assign-tag');
+    tagInput.onchange = function() {
+      makeQuery();
+    };
```
okkays, line 151: you may want to cancel previous setTimeouts so things don't come back out of order.

https://stackoverflow.com/questions/452003/cancel-kill-window-settimeout-before-it-happens

okkays (changes requested):
really nice work! One suggestion, but this pattern is going to be a _huge_ improvement to portal UX :D

crt_portal/static/js/resources.js
okkays: @kewitham I sort of expected `clearTimeout` to be called - am I missing something?

kewitham: Yep I added clearTimeout for every time setTimeout is used. There is no timeout for the tag input event listener.

okkays: ah mb - must have been looking at the outdated version - thanks for putting up with me :D

okkays (approved):

PR #1925, opened 2024-10-04
Address disposition feedback

kewitham:
[No data batches populating in 'report batches for disposal tab' #1972](https://github.com/usdoj-crt/crt-portal-management/issues/1972)
[Record rejected by default in a batch #1973](https://github.com/usdoj-crt/crt-portal-management/issues/1973)
[Default sorting for 'report batches for disposal' tab #1974](https://github.com/usdoj-crt/crt-portal-management/issues/1974)

## What does this change?
This PR addresses a few feedback items on the disposition page:
1. Deletes batches that were created but reports were not added to them
2. Updates record rejection flow to allow reports rejected from another batch to go through the flow normally in a newly created batch
3. Updated sorting on the batches page

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/views.py
```diff
@@ -715,13 +715,17 @@ def get_batch_data(disposition_batches, all_args_encoded):
 
 
 def get_batch_view_data(request):
+    # clear out any batches that were not completed
+    empty_batches = ReportDispositionBatch.objects.filter(disposed_count=0)
+    for batch in empty_batches:
+        batch.delete()
```
okkays, line 721:
```suggestion
    # clear out any batches that were not completed
    ReportDispositionBatch.objects.filter(disposed_count=0).delete()
```

it should be possible to call delete on a queryset, which will result in one query for everything instead of one for each batch

https://docs.djangoproject.com/en/5.1/ref/models/querysets/#delete

crt_portal/cts_forms/views.py
```diff
@@ -1397,6 +1401,7 @@ def get_reviewer_data(self, request, batch):
             second_reviewer_pk = None
             second_display_name = None
         return {
+            'reviewed': batch.first_reviewer is not None,
```
okkays, line 1404: nice, good idea to keep the logic out of the templates 👍

crt_portal/cts_forms/views.py
```diff
@@ -1462,6 +1467,14 @@ def post(self, request, id=None):
...
                 report_dispo_object.rejected = True
                 report_dispo_object.save()
+            approved_report_dispo_objects = ReportDisposition.objects.filter(batch=batch).exclude(public_id__in=rejected_report_ids)
+            for report_dispo_object in approved_report_dispo_objects:
+                report = Report.objects.filter(public_id=report_dispo_object.public_id).first()
+                report.batched_for_disposal = True
+                report.report_disposition_status = 'approved'
+                report.save()
+                report_dispo_object.rejected = False
+                report_dispo_object.save()
```
okkays, line 1477: it's possible there's hundreds of these in a batch, right?

that could cause problems - especially if this fails partway through. It's important to remember that `filter` doesn't actually return a list of objects - it returns a queryset. The query is only run when data on the objects is accessed. Whenever you're iterating over a queryset, look for a "filter" + "bulk update" method, instead.

For example, something like this might take this (potentially hundreds of queries) down to two:

```py
approved_report_dispo_queryset = ReportDisposition.objects.filter(batch=batch).exclude(public_id__in=rejected_report_ids)
public_ids = approved_report_dispo_queryset.values_list('public_id', flat=True)

Report.objects.filter(public_id__in=public_ids).update(report_disposition_status='approved', batched_for_disposal=True)

approved_report_dispo_queryset.update(rejected=False)
```

[stackoverflow for efficient bulk update](https://stackoverflow.com/questions/41744096/efficient-way-to-update-multiple-fields-of-django-model-object)

okkays (changes requested):

okkays (approved):
lgtm!

PR #1956, opened 2024-11-20
Add Location Zipcode Field to Organization

Rwolfe-Nava:
[Link to ZenHub issue.](https://github.com/orgs/usdoj-crt/projects/3?pane=issue&itemId=87680231&issue=usdoj-crt%7Ccrt-portal-management%7C2027)

## What does this change?

1. Adds location_zipcode as a CharField to the Report model

2. Adds location_zipcode TextField to the ReportEditForm

3. Adds location_zipcode field with usa-input class to the ReportEditForm's view

Screenshots taken from the Report form organization section in both view and edit mode:

location_zip_gif

### Author

### Reviewer

## Notes for reviewer:


crt_portal/cts_forms/forms.py
```diff
@@ -2944,6 +2946,13 @@ def __init__(self, *args, user=None, **kwargs):
         self.fields['crt_reciept_month'].widget.required = False
         self.fields['crt_reciept_year'].widget.required = False
 
+        # location fields
+        self.fields['location_zipcode'].label = LOCATION_QUESTIONS['location_zipcode']
+        self.fields['location_zipcode'].widget = TextInput(attrs={
+            'class': 'usa-input'
+        })
+        self.fields['location_zipcode'].required = False
+
```
okkays, line 2955: I was going to comment to leave this next to the other location fields, but I don't see them - is this modification of the fields variable here necessary?

okkays (changes requested):

crt_portal/cts_forms/forms.py
Rwolfe-Nava: So I think the reason we don't see them here is because the fields are coming from the ProForm. Up above, in the Meta class, we are saying:

```
        fields = ProForm.Meta.fields + ['tags', 'location_zipcode']
```

So we are bringing all of the fields in from ProForm, but then adding 'tags' and now 'location_zipcode' to the dictionary of fields. The reason I did it this way was because we don't want to add the zipcode to the normal proform, so this only adds it to the ReportEdit form.

Then, down here, we have to define the widget for the field, otherwise it will show up without the USWDS styling, and if we don't define the label here, the label defaults to 'location zipcode', which is fine, but I wanted the text to match how we do have the contact zipcode, which is "Zip code"

okkays (approved):

PR #2000, opened 2025-01-21
[Heavily WIP Still] Add Additional Contacts and Fields to Details page - CRU Proform support

Rwolfe-Nava:

## What does this change?
Refactors the way we look for the save-button on crt-toggleable-card forms (uses button--save) now instead of just looking for the first button on the form. The old way of doing it did not work if we included an element that had a button built in I.E: an accordion. This also is in line with how we were finding the 'cancel' button

Refactors the way we are configuring and grouping the additional contact fields so that we no longer need to pass in the form as an arugument. These methods have been completely rewritten and now return a nested dictionary data structure. This change was done so that we would not be required to arbitrarily recreate the entire form object on the ShowView just to get the field mapping for additional contacts.

Adds a new 'working-group' field to the Report model. This is to better represent the specific section tag we are passing through the URL. In the case of CRU proform, it is ELS-CRU. Storing this in the model allows us to recreate the necessary additional fields for these specific reports on the detail page.

Adds new form templates for use with ShowView: additional_contacts and additional_contacts_edit respectively. These are added as a card similar to complaint_details on the detail page (ShowView).

Adds a new custom django tag template for get_attribute. This is necessary because we need to retrieve fields by their attribute name on the Report object without hardcoding the actual attribute name.

Still WIP:

We need to determine business rules for 'working_group' on the detail page. Right now, working_group is only set when the record is created using the proform. This means that reassigning a record that was created with the ELS-CRU proform to a new section retains the additional_contacts fields for ELS-CRU on the record because working_group is not changed. This raises some quesitons:

If we add the ability to change the working_group, should we delete the data in these fields when working_group is changed? I would assume 'yes' since the additional contact 'kinds' will likely change, so the data in the fields will be 'bad'. The consequences of this are, we will lose this data if the report is ever reassigned back to the original working_group.

How does 'working_group' relate to 'assigned_section' if at all? Should we keep them separated with the guidance that if you change section you will need to change working_group as well?

### Author

### Reviewer

## Notes for reviewer:


okkays (approved):
lgtm! nice work!

PR #2040, opened 2025-03-21
freki-w036-6

damianj:
[Link to issue.](link-goes-here)

## What does this change?

Addresses freki-w036-6

### Author

### Reviewer

## Notes for reviewer:

### DIFF
--- a/crt_portal/api/views.py
+++ b/crt_portal/api/views.py
@@ -366,7 +366,7 @@ class ResponseList(generics.ListAPIView):
 
 
 class ResponseDetail(generics.RetrieveAPIView):
-    queryset = ResponseTemplate.objects.all()
+    queryset = ResponseTemplate.objects.filter(show_in_dropdown=True).all()
     serializer_class = ResponseTemplateSerializer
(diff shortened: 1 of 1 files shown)


crt_portal/api/views.py
```diff
@@ -366,7 +366,7 @@ class ResponseList(generics.ListAPIView):
 
 
 class ResponseDetail(generics.RetrieveAPIView):
-    queryset = ResponseTemplate.objects.all()
+    queryset = ResponseTemplate.objects.filter(show_in_dropdown=True).all()
```
damianj, line 369: Mu current understanding is that the complaint view page is the only place from where responses can be generated, if so then this change should not affect functionality as only templates enabled to be listed in the dropdown menu should come through this API endpoint via normal use.

Rwolfe-Nava, line 369: Maybe a nitpick, but I think we want to remove the `.all()` here? I don't think .all() is necessary after calling filter. I don't know if it hurts anything either, but may want to remove it just in case?

Rwolfe-Nava (approved):

